PR #15, opened 2023-07-31
update readme

ChuckHend:
Update docs for usage of partitioned queues.


samay-sharma (approved):
Overall, looks good.

I have a few unrelated suggestions to the readme that I'll propose as a PR.

Issue #20, opened 2023-08-05
CI refactor


ChuckHend:
refactor CI:
- new image published on all changes to the extension. this might either require `trunk build` in CI, or setting it up so that `trunk install` happens in the image build after `trunk publish` of the new version
- crate tests almost always fail on first run with this error:
```
Running all tests...
sqlx migrate run
error: encountered unexpected or invalid data: unexpected response from SSLRequest: 0x00
make: *** [Makefile:20: test] Error 1
```
- crate tests almost always fail on the second time due to a race condition in the tests. the rust tests all run in parallel, and several run `create table if not exist` at the same time. first time the tests there is usually a collision. subsequently, the table is created (usually pgmq_meta, I think). Might be able to just move the creation of that table into the extension/sql so that its created when `create extension` is run, rather than when the queues are first created. For the client-side-only API, that might need a jiggle to the test starts or something.
```
q_success: Err(DatabaseError(Database(PgDatabaseError { severity: Error, code: "23505", message: "duplicate key value violates unique constraint \"pg_type_typname_nsp_index\"", detail: Some("Key (typname, typnamespace)=(pgmq_meta, 2200) already exists."), hint: None, position: None, where: None, schema: Some("pg_catalog"), table: Some("pg_type"), column: None, data_type: None, constraint: Some("pg_type_typname_nsp_index"), file: Some("nbtinsert.c"), line: Some(664), routine: Some("_bt_check_unique") })))
```

PR #22, opened 2023-08-07
update CI

ChuckHend:
partially resolves https://github.com/tembo-io/pgmq/issues/20

- only publish new extensions on tag release events
- separate the Rust crate publish from tests. cargo publish will only run when there is increase to the crate version in Cargo.toml

### DIFF
--- a/.github/workflows/crate_ci.yml
+++ b/.github/workflows/crate_ci.yml
@@ -10,7 +10,7 @@ on:
     branches:
      - main
     paths:
-      - '.github/workflows/pgmq.yml'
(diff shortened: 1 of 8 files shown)


.github/workflows/crate_ci.yml
```diff
@@ -56,21 +56,3 @@ jobs:
         run: cargo sqlx prepare --check
       - name: teardown
         run: make test.cleanup
-
-  publish:
```
ChuckHend, line 60: this is now handled by `cargo_publish.yml`

nhudson (approved):

PR #23, opened 2023-08-08
add `CheckedName` as an abstraction for valid Pg identifiers

vrmiguel:
Adds `CheckedName` as a way of avoiding revalidating a Pg identifier every time

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1243,7 +1243,7 @@ name = "pgmq"
 version = "0.10.2"
 dependencies = [
  "chrono",
- "pgmq 0.14.1",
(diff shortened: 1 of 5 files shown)


ChuckHend (approved):

ChuckHend:
https://github.com/tembo-io/pgmq/actions/runs/5798457245/job/15716257108?pr=23 is going to fail because the crate version already exists in crates.io. Bumping patch on https://github.com/tembo-io/pgmq/blob/6cf5b1badd3299d8b4e066ba827c915d7f26bd61/core/Cargo.toml#L3 should resolve that.

Issue #33, opened 2023-08-09
pgmq_create doesn't validate queue names


v0idpwn:
**Extension version**: 0.10
**Postgres version**: 15.3 (I'm running your docker image)

### Current behaviour
Some characters that aren't allowed cause error 22000: 
```
>select * from pgmq_create('hello-world');
SQL Error [22000]
```
Very long names get truncated and cause issues:
```
>select * from pgmq_create('veryloooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooongstring'); 
SQL Error [42703]: column "deleted_at" does not exist
```
### Expected behaviour
If the name is valid, queue is created normally. If it isn't, error early. 

This should also probably be documented somewhere.

ChuckHend:
Nice find! Table name length limitation is 63 characters in postgres, so we probably just add that check in here: https://github.com/tembo-io/pgmq/blob/aaaf47e899dce52034f01503c60425bcc666e1d1/core/src/query.rs#L303

Thoughts?

vrmiguel:
I feel like we already handle `pgmq_create('hello-world')` but need some way of providing a better error message to the user

ChuckHend:
@vrmiguel - did #37 close this, or was there something we wanted to do before we marked this one done?

PR #37, opened 2023-08-10
Handle maximum table name length

vrmiguel:
First PR for #33 

This PR ensures tables are at most 59-characters long, since by default Postgres limits identifiers to 63 characters, and we add to it a 4-bytes long prefix. This is configurable, but would require building Postgres from source.

Docs: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS


core/src/query.rs
```diff
@@ -301,10 +301,23 @@ pub fn pop(name: &str) -> Result<String, PgmqError> {
 
 /// panics if input is invalid. otherwise does nothing.
 pub fn check_input(input: &str) -> Result<(), PgmqError> {
-    let valid = input
+    // Docs:
+    // https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS
+
+    // Default value of `NAMEDATALEN`, set in `src/include/pg_config_manual.h`
+    const NAMEDATALEN: usize = 64;
```
v0idpwn, line 308: Is there an API where this configuration can be read, so you don't need to guess?

ChuckHend: This might be close, but need to look into this a bit more. Not sure how reliable it would be for our use case.

```sql
show max_identifier_length;
 max_identifier_length 
-----------------------
 63
(1 row)
```

ChuckHend: https://www.postgresql.org/docs/current/runtime-config-preset.html

ChuckHend (approved):

PR #45, opened 2023-08-13
Remove unnecessary references

pbzweihander:
`i32` and `chrono::DateTime` implement `Copy` trait. We don't need to take reference of it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.11.2"
+version = "0.12.0"
 edition = "2021"
 authors = ["Tembo.io"]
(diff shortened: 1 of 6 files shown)


ChuckHend (approved):
Can you bump the minor version on core/Cargo.toml? Should be good to merge after that.

pbzweihander:
I have updated the version!

pbzweihander:
Oh, Should I leave outer crate's version as-is?

ChuckHend:
> Oh, Should I leave outer crate's version as-is?

We can leave the outer crate as is. I'll do a release after this merges.

Thank you for jumping in a contributing!

PR #64, opened 2023-08-17
Add pgmq_read_with_poll

v0idpwn:
For feature parity with SQS. The implementation is a bit naive: it simply will retry querying for messages on a small interval until the timeout is reached.

To be discussed: 
- should we make the interval configurable? maybe it should be a parameter to the function?
- should we receive the timeout in seconds instead of ms, for consistency, considering vt is defined in seconds?

To be done:
- add tests, i only tested manually
- add in rust client without the extension (there we could use async instead of thread::sleep)
- document

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -163,8 +163,12 @@ pub mod util;
 pub use pg_ext::PGMQueueExt;
 use util::fetch_one_message;
 
+use std::time::Duration;
+
 const VT_DEFAULT: i32 = 30;
 const READ_LIMIT_DEFAULT: i32 = 1;
+const POLL_TIMEOUT_DEFAULT: Duration = Duration::from_secs(5);
+const POLL_INTERVAL_DEFAULT: Duration = Duration::from_millis(250);
 
 /// Message struct received from the queue
 ///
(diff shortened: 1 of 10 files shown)


ChuckHend:
> * should we make the interval configurable? maybe it should be a parameter to the function?

I think we should make it configurable with a default. Postgres does let us have optional parameters in functions: https://github.com/tembo-io/pgmq/blob/df7b396ef908c892ae933b2d7b19f1cc2cb9695a/src/api.rs#L12

> * should we receive the timeout in seconds instead of ms, for consistency, considering vt is defined in seconds?

Seconds seems good to me, to be consistent with rest of pgmq. SQS is seconds too, right?

core/src/query.rs
```diff
@@ -234,14 +235,14 @@ pub fn read(name: &str, vt: i32, limit: i32) -> Result<String, PgmqError> {
         (
             SELECT msg_id
             FROM {PGMQ_SCHEMA}.{TABLE_PREFIX}_{name}
-            WHERE vt <= now()
+            WHERE vt <= clock_timestamp()
```
v0idpwn, line 238: `now()` uses transaction start time. It's pretty useful, but in this case, if we repeatedly poll it will always query with the same timestamp. I also tried `statement_timestamp()`, but for some reason while using the SPI I got the same timestamp for different runs 🤔, so I resorted to `clock_timestamp()`

core/src/lib.rs
```diff
@@ -627,6 +631,51 @@ impl PGMQueue {
...
+                }
+                Some(m) => {
+                    if m.is_empty() && start_time.elapsed() < poll_timeout_ {
+                        tokio::time::sleep(poll_interval_).await;
+                        continue;
+                    } else {
+                        break Ok(Some(m));
+                    }
+                }
+            }
```
v0idpwn, line 675: This looks awkward, accepting suggestions on how to improve it

core/src/lib.rs
```diff
@@ -627,6 +631,51 @@ impl PGMQueue {
...
+        poll_interval: Option<Duration>,
+    ) -> Result<Option<Vec<Message<T>>>, errors::PgmqError> {
+        let vt_ = vt.unwrap_or(VT_DEFAULT);
+        let poll_timeout_ = poll_timeout.unwrap_or(POLL_TIMEOUT_DEFAULT);
+        let poll_interval_ = poll_interval.unwrap_or(POLL_INTERVAL_DEFAULT);
+        let start_time = std::time::Instant::now();
+        loop {
+            let query = &query::read(queue_name, vt_, max_batch_size)?;
+            let messages = fetch_messages::<T>(query, &self.connection).await?;
+            // Why can fetch_messages return both `None` and `Option(())`
```
ChuckHend, line 657: The assumption is that if there are no messages in the queue, sqlx will return `sqlx::error::Error::RowNotFound`
https://github.com/tembo-io/pgmq/blob/df7b396ef908c892ae933b2d7b19f1cc2cb9695a/core/src/lib.rs#L891, so in that case it should be `None`. If there's any messages, then it will be `Some(Vec<Message>)`. Maybe we missed a case and somehow its returning an empty vec? Or are you saying it can the unit type?

core/src/lib.rs
ChuckHend: It's probably awkward because fetch_messages() is a little messy.

fetch_messages() is supposed to return `Some(Vec<Messages>)` if there are messages to return, otherwise it should be `None`. It shouldn't every return an empty Vec, unless there's a bug. Let me look at fetch_messages() and see if that signature can be made more explicit.

ChuckHend (approved):

ChuckHend:
@v0idpwn , the Rust client library tests are run against the container image with pgmq installed in it - and the image is currently only built when the extension is released. So the Rust client library tests will fail since they are looking for the new long poll function, which wont be released yet.

core/src/lib.rs
v0idpwn: As discussed, it was needed due to a bug in fetch_messages. Fixed bug in https://github.com/tembo-io/pgmq/pull/64/commits/a3dc314dfed8a6142f361958ec9410d55af971f8

PR #69, opened 2023-08-21
Patch assignment

ChuckHend:
assignment of the archive table to pgmq extension was adding an extra semi-colon. this made `pgmq_create()` no longer idempotent. it would fail if the queue already existed.

Also added some extra test cases.


core/src/query.rs
```diff
@@ -336,7 +336,7 @@ pub fn assign_queue(name: CheckedName<'_>) -> Result<String, PgmqError> {
 }
 
 pub fn assign_archive(name: CheckedName<'_>) -> Result<String, PgmqError> {
-    Ok(assign(&format!("{name}_archive; ")))
+    Ok(assign(&format!("{name}_archive")))
```
ChuckHend, line 339: @sjmiller609 this is the bug

sjmiller609 (approved):

PR #79, opened 2023-08-24
Add archive_batch function to ext api

v0idpwn:
- Adds variant of `pgmq_archive` that takes `bigint[]`
- Adds `archive_batch` to QueueExt

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1243,10 +1243,10 @@ dependencies = [
 
 [[package]]
 name = "pgmq"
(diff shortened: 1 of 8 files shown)


ChuckHend (approved):
🚀

v0idpwn:
I think I'm done with it. Tests fail due to crate dependency/build order issue.
I will follow this up with a PR that changes the test structure a bit, to split the testing of the ext api in more tests.

PR #87, opened 2023-08-29
Send batch

ChuckHend:
Not overloading `pgmq_send()` because it leads to ambiguity and a messy api. Implementing as a separate `pgmq_send_batch()` instead.

For example:

Valid single

```sql
pgmq=# select pgmq_send('x', '[{"hello": "world1"}, {"hello": "world0"}]'::jsonb);
 pgmq_send 
-----------
         1
```

Ambiguous, batch or single?

```sql
pgmq=# select pgmq_send('x', '[{"hello": "world1"}, {"hello": "world0"}]');
ERROR:  function pgmq_send(unknown, unknown) is not unique
LINE 1: select pgmq_send('x', '[{"hello": "world1"}, {"hello": "worl...
               ^
HINT:  Could not choose a best candidate function. You might need to add explicit type casts.
```

Valid batch
```sql
select pgmq_send('x', ARRAY['{"hello": "world1"}', '{"hello": "world0"}']);
```

Invalid batch
```sql
pgmq=# select pgmq_send('x', ARRAY['{"hello": "world1"}', '{"hello": "world0"}']);
ERROR:  function pgmq_send(unknown, text[]) does not exist
LINE 1: select pgmq_send('x', ARRAY['{"hello": "world1"}', '{"hello"...
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -5,7 +5,7 @@ use pgrx::spi::SpiTupleTable;
 use crate::errors::PgmqExtError;
 use crate::partition::PARTMAN_SCHEMA;
 use pgmq_core::{
-    query::destroy_queue,
+    query::{destroy_queue, enqueue},
     types::{PGMQ_SCHEMA, TABLE_PREFIX},
 };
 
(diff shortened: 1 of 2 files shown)


src/api.rs
```diff
@@ -69,3 +69,24 @@ pub fn listit() -> Result<Vec<(String, TimestampWithTimeZone)>, spi::Error> {
     });
     Ok(results)
 }
+
+#[pg_extern]
+fn pgmq_send_batch(
+    queue_name: &str,
+    messages: Vec<pgrx::JsonB>,
+    delay: default!(i64, 0),
```
v0idpwn, line 77:
```suggestion
    delay: default!(i64, 0)
```
I guess we will add `delay` as an optional argument for `pgmq_send` as well?

ChuckHend: Yes, thought about doing that in this PR but we can do it separately.

PR #90, opened 2023-08-29
Change deleted_at to archived_at

### LINKED ISSUE #77: [Proposal] rename `deleted_at` column to `archived_at`
This would be a breaking change, but it would be more precise. Could make it compatible by returning both archived_at and deleted_at from the functions, but it isn't great to do this forever.

v0idpwn:
Closes #77

### DIFF
--- a/sql/pgmq--0.19.0--0.20.0.sql
+++ b/sql/pgmq--0.19.0--0.20.0.sql
@@ -1,0 +1,10 @@
+DO $$
+DECLARE
+    qname TEXT;
+BEGIN
+    FOR qname IN (SELECT queue_name FROM public.pgmq_meta)
+    LOOP
+        EXECUTE format('ALTER TABLE %I RENAME COLUMN deleted_at TO archived_at', 'pgmq_' || qname || '_archive');
+        EXECUTE format('ALTER INDEX %I RENAME TO %I', 'deleted_at_idx_' || qname, 'archived_at_idx_' || qname);
+    END LOOP;
+END $$;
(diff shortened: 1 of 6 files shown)


sql/pgmq--0.19.0--0.20.0.sql
```diff
@@ -0,0 +1,10 @@
+DO $$
+DECLARE
+    qname TEXT;
+BEGIN
+    FOR qname IN (SELECT queue_name FROM public.pgmq_meta)
+    LOOP
+        EXECUTE format('ALTER TABLE %I RENAME COLUMN deleted_at TO archived_at', 'pgmq_' || qname || '_archive');
+        EXECUTE format('ALTER INDEX %I RENAME TO %I', 'deleted_at_idx_' || qname, 'archived_at_idx_' || qname);
```
v0idpwn, line 8: This index name isn't prefixed, all other things are. Could be good to take the opportunity to do so.

sql/pgmq--0.19.0--0.20.0.sql
```diff
@@ -0,0 +1,10 @@
+DO $$
```
v0idpwn, line 1: Tested this locally

sql/pgmq--0.19.0--0.20.0.sql
ChuckHend: Maybe we can leave the prefix off? Once we move everything to the pgmq schema we could drop the prefix on everything..

ChuckHend (approved):

PR #93, opened 2023-08-30
batch read in rust client

ChuckHend:


ChuckHend:
Coming back to this later.

Issue #84, opened 2023-08-28
purge queue


ChuckHend:
archive all messages from a queue

v0idpwn:
I think this can be closed now

PR #96, opened 2023-08-31
Add purge_queue function (extension, ext client)

v0idpwn:


pgmq-rs/src/lib.rs
```diff
@@ -743,6 +743,14 @@ impl PGMQueue {
         Ok(num_deleted)
     }
 
+    // TODO: release when pgmq-core 0.3.0 is released
+    //pub async fn purge(&self, queue_name: &str) -> Result<u64, PgmqError> {
+    //    let query = &core_query::purge_queue(queue_name)?;
+    //    let row = sqlx::query(query).execute(&self.connection).await?;
+    //    let num_deleted = row.rows_affected();
+    //    Ok(num_deleted)
+    //}
```
v0idpwn, line 752: Can't use new query from pgmq core

v0idpwn:
Closes #84. Notice this deletes all messages, not archive them (for equivalence with SQS API). If you want something like `archive_all` we can do it.

ChuckHend (approved):

PR #97, opened 2023-09-01
Add partitioned column to pgmq_meta

### LINKED ISSUE #91: add 'partitioned' attribute to pgmq_meta
pgmq_meta should have an attribute that tracks whether the queue is partitioned or not. This will be useful when dropping a queue, so we can also cleanup relevant entries in pg_partman. But also for general observability into existing queue.

Does not necessarily need to explicitly be "partitioned" as the attribute name, but some attribute that lets us know how the type of queue when inspecting pgmq_meta. Right now, there are just two types; partitioned and non-partitioned. In the future, there could be several types of partitioning.

craigpastro:
Resolves #91.

I had a go at resolving this.

Oh, I don't know how I should go about updating the `pgmq-rs` crate. Do you publish a new version of `pgmq` then update in a new PR?

Thanks!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1267,6 +1267,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sqlx",
+ "strum",
  "thiserror",
  "tokio",
  "url",
@@ -1650,6 +1651,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustversion"
+version = "1.0.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7ffc183a10b4478d04cbbbfc96d0873219d962dd5accaff2ffbd4ceb7df837f4"
+
+[[package]]
 name = "ryu"
 version = "1.0.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


v0idpwn (commented):
I think you'd want to provide a migration for this

craigpastro:
> I think you'd want to provide a migration for this

Thanks for your comment! Good idea. But I actually am not sure how I can do this.

One thought: I see partitioned queues have an index that non-partitioned queues do not, so I could try to list those, extract the table name from the index, and update based on that. But that sounds horrific. Maybe I am missing something.

ChuckHend:
> One thought: I see partitioned queues have an index that non-partitioned queues do not, so I could try to list those, extract the table name from the index, and update based on that. But that sounds horrific. Maybe I am missing something.

Maybe something like this? You'd also want to check first that pg_partman exists.

```
SELECT EXISTS(SELECT * FROM public.part_config where parent_table = '{queue_table}');
```

craigpastro:
Added a migration and tested locally if things are fine when pg_partman is installed and when it is not installed. Seems to work. Thanks for the hint!

ChuckHend (approved):

PR #102, opened 2023-09-03
Recreate all functions in update path

v0idpwn:
Previously, functions were added in the extension schema, but weren't added in update paths. This means that if the extension was installed in a previous version and updated with `ALTER EXTENSION`, it wouldn't include the new functions. This commit CREATES or REPLACES all functions exported by pgmq, to guarantee that users updating to pgmq 0.22 will have all functions currently exported.


ChuckHend (approved):

v0idpwn:
Merged in #104

PR #105, opened 2023-09-04
Add index to partitioned archive

ChuckHend:


v0idpwn (commented):
This looks good, but will need a migration to add it for existing queues

v0idpwn (approved):

PR #108, opened 2023-09-05
Use "generated as id" instead of "bigserial"

### LINKED ISSUE #100: Use `generated always as identity` instead of bigserial?
Here, and in other places, `bigserial` is used:
https://github.com/tembo-io/pgmq/blob/f8ae3e807c5bace58aff7da723fe3bd536aa28db/core/src/query.rs#L41

However, I think it is recommended to use `generated always as identity` over serial. See https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_serial for some reasons why.

craigpastro:
Had a go at resolving #100. 

A couple of points:
- The archive table should not have "generated as" (or bigserial) as those values will always be inserted.
- Should the underlying table of a partitioned queue have a primary key?

I tested the migration quite a bunch locally so I believe it works.

Anyway, please let me know if you think it is a worthwhile change.

Closes #100.


ChuckHend:
There is a metrics API that uses the sequence to quickly get the total number of messages that have hit the queue (pairs up well with prometheus exporter for postgres). Is there a  way to get the max `msg_id`, even when there are no messages currently in the queue?

https://github.com/tembo-io/pgmq/blob/0ffb034295fdaa47e9859d71d6032b499fcd6271/src/metrics.rs#L113

ChuckHend:
> There's some code that grants pg monitor access to the sequence. Is it required after this change? I think not, right? Due to the improved access model.

All the grants to `pg_monitor` are for metrics collection. If we move everything into the pgmq schema, then we can probably grant all to `pg_monitor` in that schema?

craigpastro:
👋 @ChuckHend. Sorry! I totally missed this conversation. I'll give it a look today. Thanks!

craigpastro:
Ok, I found a new (to me) function `pg_get_serial_sequence` that let's us do what we want. See the example below.

```
pgmq=# drop extension pgmq; create extension pgmq;
DROP EXTENSION
CREATE EXTENSION
pgmq=# select pgmq_create('foo');
 pgmq_create 
-------------
 
(1 row)

pgmq=# SELECT nextval(pg_get_serial_sequence('pgmq_foo', 'msg_id'));
 nextval 
---------
       1
(1 row)
```

There is a `GRANT` statement on a sequence also leftover. I wonder if that is needed at all anymore.

craigpastro:
Update: `nextval` may not be what we want as it increments the value. `currval` doesn't, but it is an error to call it if `nextval` has not been called (empty queue).

https://www.postgresql.org/docs/current/functions-sequence.html

craigpastro:
Sorry, one more note: the migration does not work since this will create a new sequence that starts over from 1. Need to pick up the old last value and start a new sequence with that maybe...?

craigpastro:
Another update: even with the changes in this PR, #114 still occurs.

craigpastro:
🤔 I can't get it to work with `pg_get_serial_sequence`. I will note that the current code should still work with this PR:
```
pgmq=# \dx
                                            List of installed extensions
        Name        | Version |   Schema   |                              Description                               
--------------------+---------+------------+------------------------------------------------------------------------
 pg_partman         | 4.7.3   | public     | Extension to manage partitioned tables by time or ID
 pg_stat_statements | 1.10    | public     | track planning and execution statistics of all SQL statements executed
 pgmq               | 0.24.1  | public     | A lightweight message queue. Like AWS SQS and RSMQ but on Postgres.
 plpgsql            | 1.0     | pg_catalog | PL/pgSQL procedural language
(4 rows)

pgmq=# select pgmq_create('foo');
 pgmq_create 
-------------
 
(1 row)

pgmq=# select last_value from pgmq_foo_msg_id_seq;
 last_value 
------------
          1
(1 row)
```

craigpastro:
I'll close this for now as it is terribly out-of-date.

PR #110, opened 2023-09-06
Improve pgmq read performance

v0idpwn:
WIP


src/lib.rs
```diff
@@ -126,85 +183,83 @@ fn pgmq_read(
...
 ) -> Result<
-    Vec<(
-        i64,
-        i32,
-        TimestampWithTimeZone,
-        TimestampWithTimeZone,
-        pgrx::JsonB,
-    )>,
-    spi::Error,
+    impl std::iter::Iterator<
```
v0idpwn, line 230: The core idea is that `TableIterator` accepts any iter, so we can avoid allocating and copying the data from a Vec. Instead, we can transform the iterator returned by the SPI.

PR #111, opened 2023-09-08
Very simple demo to show how benchmarking could be done with locust

ddorian:
You can run it with:
```
locust  --headless --users 10 --spawn-rate 2
```


ChuckHend (approved):

ChuckHend:
@ddorian - do you want to merge this? I think it would be good to have in the repo.

ddorian:
@ChuckHend 

Not yet. psycopg3 after 3.1.4 don't work with gevent https://github.com/psycopg/psycopg/issues/527.
So it needs to be pinned to that version. Maybe having a separate pyproject for the benchmark?

ChuckHend:
That would work. The bench doesn't need to be part of the lib that publishes to pypi.org

PR #112, opened 2023-09-08
Rewrite some core APIs in PL/pgSQL

v0idpwn:
Rewrites:
- read
- read_with_poll
- archive_batch
- archive
- delete_batch
- delete

Breaking change: instead of returning a SETOF boolean, archive_batch and delete_batch now return SETOF bigint of the message ids they deleted.

### DIFF
--- a/sql/pgmq--0.25.0--0.26.0.sql
+++ b/sql/pgmq--0.25.0--0.26.0.sql
@@ -41,7 +41,7 @@ BEGIN
             read_ct = read_ct + 1
         FROM cte
         WHERE m.msg_id = cte.msg_id
-        RETURNING *;
+        RETURNING m.msg_id, m.read_ct, m.enqueued_at, m.vt, m.message;
         $QUERY$,
         queue_name, queue_name
     );
@@ -85,7 +85,7 @@ BEGIN
               read_ct = read_ct + 1
           FROM cte
           WHERE t.msg_id=cte.msg_id
-          RETURNING *;
+          RETURNING m.msg_id, m.read_ct, m.enqueued_at, m.vt, m.message;
           $QUERY$,
           queue_name, queue_name
       );
--- a/sql/pgmq--0.25.0--0.26.0.sql
+++ b/sql/pgmq--0.25.0--0.26.0.sql
@@ -1,6 +1,8 @@
 -- Dropping rust impls
 DROP FUNCTION pgmq.read(text, integer, integer);
 DROP FUNCTION pgmq.read_with_poll(text, integer, integer, integer, integer);
+DROP FUNCTION pgmq.archive(text, bigint);
+DROP FUNCTION pgmq.archive(text, bigint[]);
 
 -- Adding message record type, required for the functions
 CREATE TYPE pgmq.message_record AS (
--- a/sql/pgmq--0.25.0--0.26.0.sql
+++ b/sql/pgmq--0.25.0--0.26.0.sql
@@ -3,6 +3,8 @@ DROP FUNCTION pgmq.read(text, integer, integer);
 DROP FUNCTION pgmq.read_with_poll(text, integer, integer, integer, integer);
 DROP FUNCTION pgmq.archive(text, bigint);
 DROP FUNCTION pgmq.archive(text, bigint[]);
+DROP FUNCTION pgmq.delete(text, bigint);
+DROP FUNCTION pgmq.delete(text, bigint[]);
 
 -- Adding message record type, required for the functions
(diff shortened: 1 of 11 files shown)


src/api.rs
```diff
@@ -505,37 +450,6 @@ mod tests {
         assert_eq!(retval.unwrap(), 1);
     }
 
-    // assert an invisible message is not readable
-    #[pg_test]
-    fn test_default() {
-        let qname = r#"test_default"#;
```
v0idpwn, line 511: I had to remove this test because it was using `pgmq_read` rust impl, but I assume this is covered in integration tests? Otherwise we can add it there.

sql/pgmq--0.25.0--0.26.0.sql
```diff
@@ -0,0 +1,104 @@
...
+            WHERE vt <= clock_timestamp()
+            ORDER BY msg_id ASC
+            LIMIT $1
+            FOR UPDATE SKIP LOCKED
+        )
+        UPDATE pgmq.queue_%s
+        SET
+            vt = clock_timestamp() + interval '$2 seconds',
+            read_ct = read_ct + 1
+        WHERE msg_id in (select msg_id from cte)
```
ChuckHend, line 40: This part of the read query changed in https://github.com/tembo-io/pgmq/pull/113

sql/pgmq--0.25.0--0.26.0.sql
```diff
@@ -0,0 +1,104 @@
...
+              WHERE vt <= clock_timestamp()
+              ORDER BY msg_id ASC
+              LIMIT $1
+              FOR UPDATE SKIP LOCKED
+          )
+          UPDATE pgmq.queue_%s
+          SET
+              vt = clock_timestamp() + interval '$2 seconds',
+              read_ct = read_ct + 1
+          WHERE msg_id in (select msg_id from cte)
```
ChuckHend, line 85:
```suggestion
          FROM cte
          WHERE t.msg_id=cte.msg_id
```

sql/pgmq--0.25.0--0.26.0.sql
```diff
@@ -0,0 +1,104 @@
...
+          WITH cte AS
+          (
+              SELECT msg_id
+              FROM pgmq.queue_%s
+              WHERE vt <= clock_timestamp()
+              ORDER BY msg_id ASC
+              LIMIT $1
+              FOR UPDATE SKIP LOCKED
+          )
+          UPDATE pgmq.queue_%s
```
ChuckHend, line 81:
```suggestion
          UPDATE pgmq.queue_%s t
```

src/sql_src.sql
```diff
@@ -14,3 +32,97 @@ BEGIN
...
+            WHERE vt <= clock_timestamp()
+            ORDER BY msg_id ASC
+            LIMIT $1
+            FOR UPDATE SKIP LOCKED
+        )
+        UPDATE pgmq.q_%s
+        SET
+            vt = clock_timestamp() + interval '$2 seconds',
+            read_ct = read_ct + 1
+        WHERE msg_id in (select msg_id from cte)
```
ChuckHend, line 65: same here as above / https://github.com/tembo-io/pgmq/pull/113/files

src/removed_functions.rs
```diff
@@ -0,0 +1,21 @@
...
+// to keep update compatibility.
+//
+#[pg_extern(sql = "")]
+pub fn pgmq_read() {}
+#[pg_extern(sql = "")]
+pub fn pgmq_read_with_poll() {}
+#[pg_extern(sql = "")]
+pub fn pgmq_archive() {}
+#[pg_extern(sql = "")]
+pub fn pgmq_archive_batch() {}
```
ChuckHend, line 21: i wonder if there is a way for us to drop these with the 1.0 release - I think some manual steps to upgrade from <1.0 -> 1.0 would be ok. I am not sure how we can get around this though.

v0idpwn: We can ask to update to 0.x (the latest 0 dot something version) and then to 1.0, and you then wouldn't need those

src/sql_src.sql
```diff
@@ -14,3 +32,205 @@ BEGIN
   END IF;
 END;
 $$ LANGUAGE plpgsql;
+
+------------------------------------------------------------
+-- Functions
+------------------------------------------------------------
```
v0idpwn, line 38: It feels pretty bad to simply copy paste those functions but I don't see a better way.

ChuckHend (approved):

PR #115, opened 2023-09-12
Migrate schemas

v0idpwn:


pgmq-rs/Cargo.toml
```diff
@@ -12,7 +12,7 @@ readme = "README.md"
 repository = "https://github.com/tembo-io/pgmq"
 
 [dependencies]
-pgmq_core = { package = "pgmq-core", version = "0.4.0" }
+pgmq_core = {package = "pgmq-core", path = "../core" }
```
ChuckHend, line 15: when users install pgmq-rs from crates.io, it will just pull down latest version of pgmq-core when compiling, right?

ChuckHend (approved):

PR #128, opened 2023-09-30
feat(ext): add is_partitioned and is_unlogged to list_queues

v0idpwn:


v0idpwn:
Extension CI fails, fix in #129

ChuckHend (approved):

Issue #54, opened 2023-08-14
Transaction support?


darky:
```sql
BEGIN;

SELECT * from pgmq_send('my_queue', '{"foo": "bar1"}');

ROLLBACK;
```

Will message exists in queue in this case? ☝️

ChuckHend:
This is definitely something we want to support and have well tested though. Thank you for bringing it up @darky!

v0idpwn:
Closing this in favor of #130. If you think additional tests should be made, we can discuss that

PR #130, opened 2023-10-01
Test transactional behaviour

v0idpwn:
We could test more (archive, delete, etc) but I don't know how far we should go since, due to our current implementation just running queries, we are pretty much testing postgres transaction isolation behaviours :)


v0idpwn:
Closes #54

ChuckHend (approved):

Issue #16, opened 2023-08-03
Report which versions of Postgres this works against (has been tested against) in the README


samay-sharma:

v0idpwn:
Closing in favor of #136 and #142

PR #136, opened 2023-10-06
Bump pgrx, postgres 16 support

v0idpwn:


vrmiguel (approved):

vrmiguel:
Needs an update to `pgx-init/action.yml` I imaginie

v0idpwn:
@vrmiguel pgx-init works out of the box (it reads pgrx version from `Cargo.toml`). There's an issue with the Extension Upgrade workflow, though: because the version of pgrx changes between 0.14.x and HEAD, we need to update `cargo-pgrx` as well.

ChuckHend:
closes https://github.com/tembo-io/pgmq/issues/16

Issue #114, opened 2023-09-10
Creating a queue with a name between 48 and 58 characters fails with an unexpected error


brianpursley:
This code that calculates the max queue name length doesn't take into consideration the sequence that is created from msg_id's bigserial type.
https://github.com/tembo-io/pgmq/blob/0ffb034295fdaa47e9859d71d6032b499fcd6271/core/src/util.rs#L107C1-L108

As a result, attempting to create a queue with a name length >= 48 and <= 58 will fail with an error due to the sequence name being truncated, instead of the "invalid queue name" error you get when the name length > 58.
```
select pgmq_create('123456789012345678901234567890123456789012345678'); -- 48 characters
```

```
[42P01] ERROR: relation "public.pgmq_123456789012345678901234567890123456789012345678_msg_id_se" does not exist Where: SQL statement "GRANT SELECT ON SEQUENCE public.pgmq_123456789012345678901234567890123456789012345678_msg_id_seq TO pg_monitor"
```

I'm guessing the name check needs to subtract 11 more to account for the `_msg_id_seq` suffix added for the sequence name, so the maximum queue name length will be 47 characters instead of 58.

pgmq version 0.24.0

v0idpwn:
Should be fixed from pgmq 0.31.1

PR #141, opened 2023-10-17
Fix calculation for maximum queue name len

v0idpwn:


tests/integration_tests.rs
```diff
@@ -111,6 +111,28 @@ async fn test_unlogged() {
...
+        .await;
+
+    assert!(result1.is_err());
+
+    let okay = 'a'.repeat(47);
+    let result2 = sqlx::query(&format!("SELECT {PGMQ_SCHEMA}.create('{okay}');"))
+        .execute(&conn)
+        .await;
+
+    assert!(result2.is_ok());
```
v0idpwn, line 133: Would be awesome if we could capture notices here somehow, as then we would be able to fail the test if postgres truncated some identifier.

v0idpwn:
Closes #114

ChuckHend (approved):

PR #153, opened 2023-11-04
add create_unlogged to python API. batch mode to benches

ChuckHend:


v0idpwn (approved):
I'd be interested in seeing the bench for more queues simultaneously. This way we can see how much we can gain by splitting work in multiple queues (which is very often the case).

Issue #30, opened 2023-08-09
Build ARM docker images


viniciusd:
I was trying to run `pgmq` on my MacBook, and I noticed `pgmq` doesn't support ARM docker images.

The current Docker manifest on quay.io is:
```$ docker manifest inspect quay.io/tembo/pgmq-pg:latest
{
	"schemaVersion": 2,
	"mediaType": "application/vnd.docker.distribution.manifest.v2+json",
	"config": {
		"mediaType": "application/vnd.docker.container.image.v1+json",
		"size": 7269,
		"digest": "sha256:3f0d3cf9de2a09d00fd4c9d2aecb1e5dbcaf20c591066fb8d42547007112f8c7"
	},
	"layers": [
		// [...]
	]
}
```

projectcalico/calico#4227 might be helpful here.

The image itself builds successfully on ARM:
```
docker build -t tembo-pgmq .                                                                                                                                                                                   [i]
[+] Building 264.9s (21/21) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                            0.0s
 => => transferring dockerfile: 2.18kB                                                                                                                                                                          0.0s
 => [internal] load .dockerignore                                                                                                                                                                               0.0s
 => => transferring context: 2B                                                                                                                                                                                 0.0s
 => [internal] load metadata for quay.io/coredb/ubuntu:22.04                                                                                                                                                    0.0s
 => [internal] load build context                                                                                                                                                                               0.0s
 => => transferring context: 52.30kB                                                                                                                                                                            0.0s
 => CACHED [builder 1/4] FROM quay.io/coredb/ubuntu:22.04                                                                                                                                                       0.0s
 => [builder 2/4] RUN apt-get update && apt-get install -y curl pkg-config libssl-dev build-essential libclang-dev clang software-properties-common                                                            56.3s
 => [builder 3/4] RUN curl https://sh.rustup.rs -sSf | sh -s -- -y                                                                                                                                             17.0s
 => [builder 4/4] RUN /root/.cargo/bin/cargo install pg-trunk                                                                                                                                                 110.1s
 => [stage-1  2/13] COPY --from=builder /root/.cargo/bin/trunk /usr/bin/trunk                                                                                                                                   0.1s
[...]

PR #160, opened 2023-11-13
build linux/amd64, linux/arm64 images

ChuckHend:
closes https://github.com/tembo-io/pgmq/issues/30


.github/workflows/extension_ci.yml
```diff
@@ -165,23 +163,46 @@ jobs:
...
-          docker push quay.io/tembo/pgmq-pg:latest
+
+      - name: Build and push -- Commit
+        # push commit build when not a release
+        if: github.event_name != 'release'
+        uses: docker/build-push-action@v5
+        with:
+          file: ./images/pgmq-pg/Dockerfile
+          context: .
+          platforms: linux/amd64, linux/arm64
```
ChuckHend, line 192: maybe only build amd64 on each commit? emulated build takes a long time...

v0idpwn: Have you considered native ARM runners? I know github actions itself currently doesn't supply those, but third parties do and its not too hard to set up self hosted runners. If the emulated build is taking too long, could be a decent option.

ChuckHend: yeah that would ideal!

v0idpwn (approved):

PR #181, opened 2024-01-11
Add PGXN release workflow

theory:
PGXN requires a `META.json` file, so add it and a `make` target to fill in the version from `Cargo.toml`. Add `.gitattributes` to prevent `git archive` from archiving some files (should more be omitted from the release? What about the `.cargo` directory?) and
`.github/workflows/pgxn-release.yml` to do the actual release when a tag is pushed.

Also tweak the `README` a bit to address some inconsistencies, and to include a PGXN version badge.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+.gitignore export-ignore
+.gitattributes export-ignore
+.github export-ignore
+*.bak export-ignore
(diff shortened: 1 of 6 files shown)


theory:
[Build is good](https://github.com/tembo-io/pgmq/actions/runs/7493599513/job/20399649019). Once the PR is approved, I'll change `ls -lah` to `pgxn-release`. Once it releases, I'll remove the `branches: [pgxn]` line, leaving releases to happen when tags are pushed starting with `v`.

ChuckHend (approved):

PR #186, opened 2024-01-27
pgmq-core - make the `enqueue` method return a parametrized query

nimrodkor:
Part #1 of 2 for #185 - this enhnaces the core so it will generate a parametrized query allowing sqlx / pg to do the sanitization of the input

PR #187, opened 2024-01-27
pgmq-rs - Leverage parametrized queries from core

### LINKED ISSUE #185: pgmq-rs can't save strings with apostrophe in them
Hey guys!

This is a great crate, thanks so much!

I'm trying to use it to encode a message, which is a string we get from the user, and apparently if that string includes an apostrophe it fails to save because, well, the lib doesn't escape it. Example:
let's say I want to write the following JSON message:
```json
{
     "to-do": "don't delete before you read it"
}
```

Then, when I try to save it, it turns into the following query:
```sql
INSERT INTO pgmq.q_github_events_jobs (vt, message)
VALUES ((now() + interval '0 seconds'), '{
    "to-do": "don't delete before you read it"
}'::json)
	        RETURNING msg_id;
```

Which even this preview shows is broken. Need to escape the `'` characters by duplicating them, i.e. this would work:
```sql
INSERT INTO pgmq.q_github_events_jobs (vt, message)
VALUES ((now() + interval '0 seconds'), '{
    "to-do": "don''t delete before you read it"
}'::json)
	        RETURNING msg_id;
```

Are you guys open for a contribution? Any specific pointers?

nimrodkor:
Part 2 of 2. Makes pgmq-rs leverage the parametrized methods in the new core version that will be published.

Will break UTs until the new version is published

Depends on #186 . Resolves #185

### DIFF
--- a/pgmq-rs/src/lib.rs
+++ b/pgmq-rs/src/lib.rs
@@ -440,7 +440,8 @@ impl PGMQueue {
         messages: &[T],
     ) -> Result<Vec<i64>, PgmqError> {
         let mut msg_ids: Vec<i64> = Vec::new();
-        let mut q = sqlx::query(&core_query::enqueue(queue_name, messages.len(), &0)?);
+        let query = core_query::enqueue(queue_name, messages.len(), &0)?;
+        let mut q = sqlx::query(&query);
         for msg in messages.iter() {
             q = q.bind(serde_json::json!(msg));
         }
--- a/pgmq-rs/src/lib.rs
+++ b/pgmq-rs/src/lib.rs
@@ -320,8 +320,8 @@ impl PGMQueue {
         message: &T,
     ) -> Result<i64, PgmqError> {
         let msg = serde_json::json!(&message);
-        let msgs: [serde_json::Value; 1] = [msg];
-        let row: PgRow = sqlx::query(&core_query::enqueue(queue_name, &msgs, &0)?)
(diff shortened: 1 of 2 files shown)


pgmq-rs/src/lib.rs
```diff
@@ -440,15 +439,12 @@ impl PGMQueue {
         queue_name: &str,
         messages: &[T],
     ) -> Result<Vec<i64>, PgmqError> {
-        let mut msgs: Vec<serde_json::Value> = Vec::new();
         let mut msg_ids: Vec<i64> = Vec::new();
+        let mut q = sqlx::query(&core_query::enqueue(queue_name, messages.len(), &0)?);
```
ChuckHend, line 443: might have to move `core_query::enqueue(queue_name, messages.len(), &0)` into its own let statement to resolve:

```
   --> src/lib.rs:443:34
    |
443 |         let mut q = sqlx::query(&core_query::enqueue(queue_name, messages.len(), &0)?);
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ - temporary value is freed at the end of this statement
    |                                  |
    |                                  creates a temporary value which is freed while still in use

```

ChuckHend: I pushed a change up for this. Turns out that change we merged in `pgmq-core` was breaking, so we need to get this pushed out now to resolve it. I should have caught that before, maybe we can get a test in place for it.

ChuckHend (approved):
Thank you for the contributions!

PR #201, opened 2024-03-08
Set minimum psycopg version to 3.1.14 which includes support for gevent (for locust load testing)

ddorian:
Support was added in 3.1.14 https://www.psycopg.org/psycopg3/docs/news.html#psycopg-3-1-14.


ChuckHend:
Thank you @ddorian, good call. Can we move that under the dev dependencies section in pyproject.toml? Or we could create a new "benchmarking" section in there, but I think just dev is fine. Users just importing the library wont need gevent compatibility, only folks trying to run the benchmarks.

ddorian:
Now it's added only in dev.

ChuckHend (approved):

ChuckHend:
Thank you @ddorian !

Issue #199, opened 2024-03-05
Any interest in a pgrx-ectomy?


olirice:
Hi from Supabase 👋,

I've been looking into queuing solutions and noticed pgmq depends pgrx, but that usage is mostly for SQL templating and SPI.

Curious if I missed any features that are heavily dependent on pgrx/rust or if this could potentially be plain plpgsql extension (and  [TLE](https://github.com/aws/pg_tle))?

If so, would there be interest in PRs pushing this project in that direction?

v0idpwn:
The only problem with such a transition is that it would deter further optimization (such as what is discussed in issue #70)

ChuckHend:
I think there could be a distribution that's compatible with TLE, basically package everything that can be moved to the SQL files and ship it. If there are shared objects that aren't compatible with TLE then those just don't ship to TLE.

ChuckHend:
Now that #212 has merged, I think we just need to migrate the build system off `pgrx`. Probably makes sense to use [pgxs](https://www.postgresql.org/docs/current/extend-pgxs.html)to build now...

v0idpwn:
Done in #230. Will be released soon on 1.3.2

ChuckHend:
Closing this now. I don't think there is anything blocking pgmq from TLE and dbdev.

PR #212, opened 2024-05-16
Implement all API functions in PL/pgSQL

v0idpwn:
Progress for #199


v0idpwn:
@ChuckHend I'm getting `PRIMARY KEY constraint on table \"q_test_duration_804\" lacks column \"enqueued_at\" which is part of the partition key.` for the case where the partition column is `enqueued_at`. It seems to me that I've the exact same queries but there must be some oversight. Do you have any idea of what could be causing this?

v0idpwn:
Fixed the `is_partitioned` issue. 

I'm able to reproduce it locally, it only happens when you use a time-based partition key, for example:

```
pgmq=# select * from pgmq.create_partitioned('asd9', 'daily', 'weekly');
ERROR:  unique constraint on partitioned table must include all partitioning columns
DETAIL:  PRIMARY KEY constraint on table "q_asd9" lacks column "enqueued_at" which is part of the partition key.
```

ChuckHend:
Strange, I can reproduce it now too. Also, executing the `CREATE TABLE` statement that is shown in the error directly also produces the same error.

```
pgmq=# select pgmq.create_partitioned('part_2', 'daily', 'weekly');
ERROR:  unique constraint on partitioned table must include all partitioning columns
DETAIL:  PRIMARY KEY constraint on table "q_part_2" lacks column "enqueued_at" which is part of the partition key.
CONTEXT:  SQL statement "
        CREATE TABLE IF NOT EXISTS pgmq.q_part_2 (
            msg_id BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
            read_ct INT DEFAULT 0 NOT NULL,
            enqueued_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
            vt TIMESTAMP WITH TIME ZONE NOT NULL,
            message JSONB
        ) PARTITION BY RANGE (enqueued_at)
    "
PL/pgSQL function pgmq.create_partitioned(text,text,text) line 9 at EXECUTE
```

ChuckHend:
I wonder if this is has anything to do with creating the partitioned index within the same transaction that the table is created?

ChuckHend:
@v0idpwn - the current create statement for partition queue does not make `msg_id` a primary key during `create table`

https://github.com/tembo-io/pgmq/blob/b214772c22f69411fd672ba9e13fd256f56142ee/src/partition.rs#L82

Removing the pkey from the create statement seems to let make this work. I wonder if this is a bug on `main` branch though, and we should be running something like this when partitioning by a time range.

```sql
CREATE TABLE IF NOT EXISTS pgmq.q_myqueue (
    msg_id BIGINT GENERATED ALWAYS AS IDENTITY,
    read_ct INT DEFAULT 0 NOT NULL,
    enqueued_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
    vt TIMESTAMP WITH TIME ZONE NOT NULL,
    message JSONB,
    PRIMARY KEY (enqueued_at, msg_id) <-- new
) PARTITION BY RANGE (enqueued_at);
```

v0idpwn:
This indeed works! I didn't see the difference. Thanks!

v0idpwn:
I think this is complete.
There's further work to be done but it could be done in future PRs.

vrmiguel (approved):

ChuckHend (approved):
Great work @v0idpwn

src/sql_src.sql
```diff
@@ -337,3 +344,433 @@ BEGIN
...
+        num := partition_interval::INTEGER;
+        RETURN 'msg_id';
+    EXCEPTION
+        WHEN others THEN
+            RETURN 'enqueued_at';
+    END;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE FUNCTION pgmq.ensure_pg_partman_installed()
```
ChuckHend, line 659: should this maybe be `_ensure_pg_partman_installed()` similar to `_get_partitioned_col()`?

v0idpwn: I think it's a good idea.

v0idpwn: Done in https://github.com/tembo-io/pgmq/pull/212/commits/70121647174367d35e409b72d17108701be0b1dc

PR #219, opened 2024-05-21
update sqlx cache in pgmq-rs

ChuckHend:
v1.2.0 of the extension changed the order of the columns returned by `pgmq.list_queues()`. I like the new way better, and is only an issue for the sqlx cache for the Rust client. Quick fix in this PR.

old:

```text
 queue_name | is_partitioned | is_unlogged |          created_at           
------------+----------------+-------------+-------------------------------
 x          | f              | f           | 2024-05-21 18:53:45.589962+00
```

new:

```text
 queue_name |          created_at           | is_partitioned | is_unlogged 
------------+-------------------------------+----------------+-------------
 x          | 2024-05-21 18:56:11.791476+00 | f              | f
 ```


ChuckHend:
related https://github.com/tembo-io/pgmq/pull/217

v0idpwn (approved):
Maybe this should be considered a breaking change from the extension (hence, a bug) and fixed in 1.2.1?

Approving in case you disagree

PR #222, opened 2024-05-23
feat: Add environment variable support to PGMQueue and update tests

tavallaie:
### Description

This pull request adds environment variable support to the `PGMQueue` class and updates the tests to reflect these changes. The environment variables are loaded from a `.env` file using `python-dotenv`. This allows for easier configuration and setup of the database connection parameters.

### Changes Made

- Added `from_env` class method to `PGMQueue` to initialize settings from a `.env` file.
- Integrated `dotenv` to load environment variables.
- Updated `test_integration.py` to use the new `PGMQueue` initialization method.
- Removed unnecessary test files (`tests/conftest.py` and `tests/test_queue.py`) to clean up the test suite.
- Simplified `pyproject.toml` by removing unused dependencies and adding `python-dotenv`.
- Updated `README.md` to include instructions for setting up and using environment variables.

### How to Test

1. Ensure you have a `.env` file in the root directory with the following content:

    ```
    PG_HOST=127.0.0.1
    PG_PORT=5432
    PG_USERNAME=postgres
    PG_PASSWORD=postgres
    PG_DATABASE=test_db
    ```

2. Run the tests:

    ```sh
    python -m unittest test_integration.py    
    ```


tembo-pgmq-python/tembo_pgmq_python/queue.py
```diff
@@ -47,99 +48,101 @@ def __post_init__(self) -> None:
         with self.pool.connection() as conn:
             conn.execute("create extension if not exists pgmq cascade;")
 
+    @classmethod
+    def from_env(cls, env_file=".env"):
```
v0idpwn, line 52: I'm not a fan of this function.

I don't think libraries should load environment files automatically, but **at most** read environment variables, and even that is questionable IMO.

tembo-pgmq-python/tests/.env
```diff
@@ -0,0 +1,5 @@
+PG_HOST=foolishly-mellow-cottonmouth.data-1.use1.tembo.io
+PG_PORT=5432
+PG_USERNAME=postgres
+PG_PASSWORD=qNr4LAz4FAVzRNNj
+PG_DATABASE=postgres
```
v0idpwn, line 5: I'm guessing this was unintentionally added? Should probably use a local database instance for tests, etc

tavallaie: Yes it was unintentional I'll fix it.

tembo-pgmq-python/tembo_pgmq_python/queue.py
tavallaie: I'm aiming to make the library as easy to use as possible. I added this function because I encountered some issues with pydantic-settings. If you're not satisfied with it, we can remove it. My goal was to improve testing

v0idpwn: I really like the testing changes, but this part strikes me as "not right". I'm not a python expert, though, so let's wait and see what others think :)

tembo-pgmq-python/pyproject.toml
```diff
@@ -15,29 +15,12 @@ packages = [{include = "tembo_pgmq_python"}]
...
-sqlalchemy = "^2.0.13"
-psycopg2 = "^2.9.6"
-scipy = "^1"
-typer = {extras = ["all"], version = "^0.9.0"}
-types-psycopg2 = "^2.9.21.11"
-locust = "^2.16.1"
 ruff = "^0.3.1"
-pyyaml = "^6.0.1"
-seaborn = "^0.13.0"
 psycopg = {version = "^3.1.14", extras = ["binary", "pool"]}
```
ChuckHend, line 24: Was it intentional to remove all of the dev dependencies? the [benchmark scripts](https://github.com/tembo-io/pgmq/tree/main/tembo-pgmq-python/benches) depend on most of this stuff. The intention of them being in dev group is so that they are not required to run the main application.

tembo-pgmq-python/tembo_pgmq_python/queue.py
ChuckHend: I really like how this adds the ability to read the values from environment variables. I think I would prefer to have it skip the direct call to python-dotenv (let the user do that). Instead, if user provides values directly through arguments, use them, otherwise fall back to the environment variables (the names you chose are good). If the values weren't provided in either env var or arguments, then it should fail.

tembo-pgmq-python/tembo_pgmq_python/queue.py
```diff
@@ -47,99 +48,101 @@ def __post_init__(self) -> None:
...
-            A value of 10 will create new/delete partitions every 10 seconds. This value should be tuned
-            according to the volume of messages being sent to the queue.
-
-        Args:
-            queue: The name of the queue.
-            partition_interval: The number of messages per partition. Defaults to 10,000.
-            retention_interval: The number of messages to retain. Messages exceeding this number will be dropped.
-                Defaults to 100,000.
-        """
-
```
ChuckHend, line 66: Is there another place that this comment could live? This is important information for users of partitioned queues. I think that documentation for the arguments are useful, but also could maybe live somewhere else if it makes more sense.

tembo-pgmq-python/Makefile
```diff
@@ -16,4 +16,4 @@ run-pgmq-postgres:
 	docker run -d --name pgmq-postgres -e POSTGRES_PASSWORD=postgres -p 5432:5432 quay.io/tembo/pgmq-pg:latest || true
 
 test: run-pgmq-postgres
-	poetry run pytest
+	poetry run python -m unittest discover tests
```
ChuckHend, line 19: `pytest` vs `unittest` is certainly a debate. Curious to hear your thoughts on why `unittest` over `pytest` for this project.

tembo-pgmq-python/tembo_pgmq_python/queue.py
tavallaie: I was thinking of adding it to the README, but I forgot. If it's very important information, I believe it should be included here as a docstring.

tembo-pgmq-python/Makefile
tavallaie: The main reason I chose `unittest` is because it has fewer dependencies. Since `unittest` is part of the Python standard library, we don't need to install anything extra. Another reason is that with new versions of Python, like 3.13 and above, there are many changes, such as removing the GIL and using JIT (Just In Time) compilation. We should wait for `pytest` to update and support these new features. Additionally, `unittest` has good test discovery built-in, which helps us easily find and run our tests.

ChuckHend: Thank you. 

> Another reason is that with new versions of Python, like 3.13 and above, there are many changes, such as removing the GIL and using JIT (Just In Time) compilation.

I agree that there are new versions of python being released, but I'm not sure how pytest relates. Does `pytest` not run on python 3.13? 

Overall though, I don't think there is anything wrong with pytest, it is a more complete test framework than unittest.  That being said, I'm ok with moving to `unittest` if that means we get more complete test coverage on the project.

ChuckHend: If we are going to change the test framework to `unittest` in this PR, then we'll also need to update the [Makefile](https://github.com/tembo-io/pgmq/blob/6f3e1e5af9c81be323550908d2994fc625f191f3/tembo-pgmq-python/Makefile#L18-L19). This is called in the [CI environment](https://github.com/tembo-io/pgmq/blob/6f3e1e5af9c81be323550908d2994fc625f191f3/.github/workflows/pgmq_python.yml#L52-L53)

tavallaie: I generally prefer to avoid adding external dependencies. pytest is an external package, while unittest is built into Python and does not require installation. Since we are not using any pytest plugins, I believe it's best to keep things simple and stick with unittest unless pytest becomes absolutely necessary.

tembo-pgmq-python/pyproject.toml
tavallaie: I added the minimum development dependencies to the bench group.

tembo-pgmq-python/tembo_pgmq_python/queue.py
tavallaie: so what do you think I should change?

tembo-pgmq-python/Makefile
ChuckHend: Thank you. 🤦 my bad!

tembo-pgmq-python/tembo_pgmq_python/queue.py
ChuckHend: I think this would be very nice:

1. remove `from_env()` and the dependency on python-dotenv
2. expand on `__post_init__` to pull in values from environment variables (similar to how you have it in this PR) 
3. if any of the attributes are provided as arguments, override the env variables with the argument values
4. if we dont have enough information to produce a valid connection string, either fall back to the default values (localhost:5432/postgres) or raise an exception I would be ok going either way with this (use defaults, or require user to provide connection info). @v0idpwn any thoughts on this?

v0idpwn: Agreed!

tembo-pgmq-python/pyproject.toml
ChuckHend: Thank you! We will need to update `poetry.lock` as well since the dependencies are getting shuffled around. I think `poetry lock --no-update` should work, but I think it will give you an error that you can follow if you run `poetry install` with a new environment.

tembo-pgmq-python/tembo_pgmq_python/queue.py
tavallaie: done! but I learn a lot, painfully. :)))))))

tembo-pgmq-python/README.md
```diff
@@ -20,23 +20,56 @@ Postgres running the [Tembo PGMQ extension](https://github.com/tembo-io/tembo/tr
 docker run -d --name postgres -e POSTGRES_PASSWORD=postgres -p 5432:5432 quay.io/tembo/pgmq-pg:latest
 ```
 
-### Initialize a connection to Postgres
+### Using Environment Variables
+
+Create a `.env` file in your project root with the following content:
```
ChuckHend, line 25: I don't think we should use python-dotenv in the readme example since we are going for simplicity and minimizing external dependencies. Can we do something like this instead?

Having trouble formatting the nested code block, so apologies for that.

--- 

Set environment variables

```text
export PG_HOST=127.0.0.1
export PG_PORT=5432
export PG_USERNAME=postgres
export PG_PASSWORD=postgres
export PG_DATABASE=test_db
```

PGMQ initialize a connection using the environment variables by default:

```python
from tembo_pgmq_python import PGMQueue, Message

queue = PGMQueue()
```

tavallaie: Ok, all do that. I also remove `python-dotenv` from dependencies. but for test of this differences I think we need more details to implement.

v0idpwn (approved):

ChuckHend (approved):

ChuckHend (approved):
Awesome! We can bump the version in pyproject.toml whenever we're want to release a new package to pypi.

tavallaie:
For new version, wait to add some real features.

Issue #221, opened 2024-05-22
With pgmq 1.2.0, pop returns an empty row when called on an empty queue


brianpursley:
The behavior of pop() is different in pgmq 1.2.0 than it is in 1.1.1.

Using this to test:
```
SELECT pgmq.create('my_queue');
SELECT msg_id, read_ct, enqueued_at, vt, message FROM pgmq.pop('my_queue');
```

Output with pgmq 1.1.1
```
npgmq_test.public> SELECT pgmq.create('my_queue')
[2024-05-21 21:20:34] 1 row retrieved starting from 1 in 39 ms (execution: 27 ms, fetching: 12 ms)
npgmq_test.public> SELECT msg_id, read_ct, enqueued_at, vt, message FROM pgmq.pop('my_queue')
[2024-05-21 21:20:34] 0 rows retrieved in 19 ms (execution: 5 ms, fetching: 14 ms)
```

Output with pgmq 1.2.0
```
npgmq_test.public> SELECT pgmq.create('my_queue')
[2024-05-21 21:21:16] 1 row retrieved starting from 1 in 37 ms (execution: 24 ms, fetching: 13 ms)
npgmq_test.public> SELECT msg_id, read_ct, enqueued_at, vt, message FROM pgmq.pop('my_queue')
[2024-05-21 21:21:17] 1 row retrieved starting from 1 in 22 ms (execution: 5 ms, fetching: 17 ms)
```

Notice `1 row retrieved` (all values in the row are null).

I also tested with read(), and it does not return a row when the queue is empty, as expected. So this seems to only be a difference in how pop() works.

v0idpwn:
The same may be happening for set_vt.

ChuckHend:
I think these need  to return `SETOF` and `RETURN QUERY `. Like how [pgmq.read()](https://github.com/tembo-io/pgmq/blob/46446995a3ec32bf9f7949ebd83cdb10c3a307a1/src/sql_src.sql#L49)  is implemented.

```sql
CREATE FUNCTION pgmq.pop(queue_name TEXT)
RETURNS SETOF pgmq.message_record AS $$
```
...and

```sql
RETURN QUERY EXECUTE sql;
```

ChuckHend:
@brianpursley , we patched this in https://github.com/tembo-io/pgmq/pull/223 and it has now been released. Thank you for reporting the bug and let us know if you see anything else!

PR #223, opened 2024-05-23
use set of pop/set_vt

ChuckHend:
Closes https://github.com/tembo-io/pgmq/issues/221


v0idpwn (approved):
LGTM, just needs the migration

ChuckHend:
@v0idpwn , I added migration + some tests for set_vt and pop

v0idpwn (approved):

Issue #216, opened 2024-05-16
Enhancements for PGMQueue Python Library


tavallaie:
I am considering adding new functionalities to the PGMQueue library to enhance its capabilities. Below is a list of potential functions. I am seeking input from other developers on which functions should be prioritized, and any additional suggestions.

### Potential New Functions

1. **Queue Status/Length:**
    - `get_queue_length`: Get the current number of messages in the queue.

2. **Queue Monitoring:**
    - `get_queue_stats`: Retrieve various statistics about the queue, such as the number of messages enqueued, processed, failed, etc.

3. **Message Management:**
    - `requeue_message`: Requeue a specific message to process it again.
    - `update_message`: Update the content of a specific message in the queue.

4. **Batch Operations:**
    - `delete_batch`: Delete a batch of messages by their IDs.
    - `archive_batch`: Archive a batch of messages by their IDs.

5. **Advanced Queue Operations:**
    - `pause_queue`: Temporarily pause processing of messages in the queue. **(Requires schema changes)**
    - `resume_queue`: Resume processing of messages in a paused queue. **(Requires schema changes)**
    - `delete_queue`: Delete an entire queue and all its messages. **(May require schema changes)**

6. **Message Retrieval:**
    - `peek_message`: Retrieve a message without removing it from the queue.
    - `peek_batch`: Retrieve a batch of messages without removing them from the queue.

### How You Can Help

Please provide feedback on the following:
1. Which functions do you think are most critical to add?
2. Are there any additional functions you believe would be beneficial?
3. Any other suggestions or comments?

ChuckHend:
Great ideas @tavallaie . Here's some thoughts:

For all of these features that already have a corresponding SQL function, I think it makes a lot of sense to just implement the function in the python lib. `pgmq.metrics()`, `pgmq.metrics_all()` have queue length, and some other stats about the queue. `pgmq.delete()` and `pgmq.archive()` both support batch, but this is not implemented in the python lib yet.

I like the requeue idea, and we can probably implement that using `pgmq.set_vt()`. Also update existing message would be useful. A separate feature people have asked for is something like `pgmq.delete_where()` that would delete all messages in a queue that have messages containing some value.

Pause and resume would be nice features too. Are you thinking this would prevent new messages from being able to reach the queue, or just prevent existing messages from being consumed?

tavallaie:
I'm sorry for being late :)
I was occupied with PR #222.
[...]

PR #227, opened 2024-05-23
add support for metrics to python lib

tavallaie:
add support for `metrics` and `metrics_all`  due to #216 with test abd documentations


ChuckHend (approved):
Excellent 😸

Issue #137, opened 2023-10-07
build images for all supported pg versions


ChuckHend:
Publish docker images for all the versions that we run tests for / are supported.

tavallaie:
I thinks with docker like this: 
``` yaml
ARG PG_VERSION
ARG PGRX_VERSION

FROM quay.io/coredb/pgrx-builder:pg${PG_VERSION}-pgrx${PGRX_VERSION}

WORKDIR /app

COPY --chown=postgres:postgres . .

ARG EXTENSION_NAME
ARG EXTENSION_VERSION

RUN cd extension && cargo pgrx package

# Add commands to build and push Docker image
FROM postgres:${PG_VERSION}

COPY --from=0 /app/target/release/${EXTENSION_NAME}-${EXTENSION_VERSION}.so /usr/lib/postgresql/${PG_VERSION}/lib/

ENTRYPOINT ["postgres"]

```

and action like this: 
```yaml
name: Build and Push Docker Images

on:
  push:
    tags:
      - 'v*.*.*'

jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        pg_version: [13, 14, 15]  # Add all supported PG versions
        pgrx_version: [0.9.8]     # Add relevant PGRX versions

    steps:
    - name: Checkout repository
      uses: actions/checkout@v2

    - name: Set up Docker Buildx
      uses: docker/setup-buildx-action@v1

    - name: Log in to Docker Hub
      uses: docker/login-action@v1
      with:
        username: ${{ secrets.DOCKER_USERNAME }}
        password: ${{ secrets.DOCKER_PASSWORD }}

    - name: Build and push
      uses: docker/build-push-action@v2
      with:
        context: .
        push: true
        tags: |
          yourdockerhubusername/pgmq:${{ matrix.pg_version }}-${{ matrix.pgrx_version }}
          ${{ steps.set-latest.outputs.latest_tag }}
      build-args: |
        PG_VERSION=${{ matrix.pg_version }}
        PGRX_VERSION=${{ matrix.pgrx_version }}

    - name: Set latest tag
      id: set-latest
      run: |
        PG_VERSIONS=(${{ matrix.pg_version }})
        HIGHEST_PG_VERSION=$(echo ${PG_VERSIONS[*]} | tr ' ' '\n' | sort -nr | head -n1)
        if [ "${{ matrix.pg_version }}" == "$HIGHEST_PG_VERSION" ]; then
[...]

PR #233, opened 2024-05-26
build images for all supported pg versions #137

tavallaie:
change about build images for all supported pg versions #137


.github/workflows/extension_ci.yml
```diff
@@ -151,25 +152,24 @@ jobs:
           trunk publish pgmq --version ${pgmq_ver} --description "A lightweight message queue. Like AWS SQS and RSMQ, on Postgres." --homepage "https://github.com/tembo-io/pgmq" --repository "https://github.com/tembo-io/pgmq" --license "PostgreSQL" --category featured --category orchestration
   build_and_push:
     name: Build and push images
-    # needs:
-    #   - publish
-    runs-on:
-      - self-hosted
-      - dind
-      - large-8x8
```
ChuckHend, line 159: pgrx can take quite a bit of resources. I think we'll want to use the self hosted runners for this.

.github/workflows/extension_ci.yml
```diff
@@ -190,26 +190,24 @@ jobs:
           password: ${{ secrets.QUAY_PASSWORD_TEMBO }}
 
       - name: Build and push -- Commit
-        # push commit build when not a release
         if: github.event_name != 'release'
         uses: docker/build-push-action@v5
         with:
-          file: ./images/pgmq-pg/Dockerfile
+          file: ./images/pgmq-${{ matrix.pg_version }}/Dockerfile
```
ChuckHend, line 196: Does this line need to change? There is no dir `./images/pgmq-pg16` for example and we should be able to use a single dockerfile, right?

images/pgmq-pg/Dockerfile
```diff
@@ -29,18 +30,18 @@ RUN $HOME/.cargo/bin/cargo pgrx install --pg-config=$(which pg_config)
...
-COPY --from=builder /usr/lib/postgresql/15/lib /usr/lib/postgresql/15/lib
+COPY --from=builder /usr/share/postgresql/${PG_VERSION#pg}/extension /usr/share/postgresql/${PG_VERSION#pg}/extension
+COPY --from=builder /usr/lib/postgresql/${PG_VERSION#pg}/lib /usr/lib/postgresql/${PG_VERSION#pg}/lib
 
 RUN apt-get update \
-    && apt-get install -y ca-certificates
+	&& apt-get install -y ca-certificates
 
-COPY ./images/pgmq-pg/postgresql.conf /usr/share/postgresql/15/postgresql.conf.sample
+COPY ./images/pgmq-${PG_VERSION}/postgresql.conf /usr/share/postgresql/${PG_VERSION#pg}/postgresql.conf.sample
```
ChuckHend, line 44: Maybe?

```suggestion
COPY ./images/pgmq-pg/postgresql.conf /usr/share/postgresql/${PG_VERSION#pg}/postgresql.conf.sample
```

ChuckHend:
Thank you for picking this up!

.github/workflows/extension_ci.yml
ChuckHend: once we are completely moved off pgrx then this should no longer be a concern

tavallaie: I am not familiar with your build system yet so some times I miss configurations.

ChuckHend:
I need to figure out why [docker login](https://github.com/tembo-io/pgmq/actions/runs/9244770049/job/25430450379?pr=233) is not working properly on this PR.

v0idpwn:
Probably because it's running from a fork without the secrets? Sec...

v0idpwn:
Yep, I created #235 and it passes this part but fails for other reasons.

v0idpwn:
@tavallaie for testing, I think you can comment out the login step. Then it will try to build and fail at pushing. Once it fails at pushing only, we can merge :)

After merge, we can probably disable auth/push for PRs, these should run only on main branch.

tavallaie:
it seems logical to me, contributor should not have such access someone can extract secret from main repositories :)

tavallaie:
@v0idpwn I think I messed up somewhere in action and I need help.

v0idpwn:
@tavallaie you added the input but you need to set it as an environment variable when building the docker image

tavallaie:
it try to get postgresql-server-dev-15.7 but the last version I can find is [15.6-0+deb12u1](https://packages.debian.org/search?keywords=postgresql-server-dev-15) so how can prevent that?

.github/workflows/extension_ci.yml
```diff
@@ -151,25 +152,27 @@ jobs:
...
   build_and_push:
     name: Build and push images
-    # needs:
-    #   - publish
     runs-on:
       - self-hosted
       - dind
       - large-8x8
+    strategy:
+      matrix:
```
v0idpwn, line 160:
```suggestion
      fail-fast: false
      matrix:
```
I'm interested in how it behaves with other versions

v0idpwn:
@tavallaie The image is overriding PG_VERSION: https://github.com/docker-library/postgres/blob/8a0b96710d917d1c3b32a5fe5b66687ad83827da/15/bookworm/Dockerfile#L93

Maybe use a different name?

images/pgmq-pg/Dockerfile
```diff
@@ -1,9 +1,10 @@
-FROM postgres:15.4-bookworm as builder
+# syntax=docker/dockerfile:1.4
 
-ARG PGMQ_VER=0.33.1
+ARG PG_DOCKER_VERSION
+FROM postgres:${PG_DOCKER_VERSION}-bookworm as builder
```
v0idpwn, line 4: Need to redeclare the argument to keep using it afterwards, see https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact

images/pgmq-pg/Dockerfile
```diff
@@ -29,18 +31,19 @@ RUN $HOME/.cargo/bin/cargo pgrx install --pg-config=$(which pg_config)
...
 # install pg_partman
 RUN git clone https://github.com/pgpartman/pg_partman.git && \
-    cd pg_partman && \
-    git checkout v4.7.4 && make && make install
+	cd pg_partman && \
+	git checkout v4.7.4 && make && make install
 
-FROM postgres:15.4-bookworm
+FROM postgres:${PG_DOCKER_VERSION}-bookworm
+ARG PG_DOCKER_VERSION
```
v0idpwn, line 38:
```suggestion
ARG PG_DOCKER_VERSION=$PG_DOCKER_VERSION
FROM postgres:${PG_DOCKER_VERSION}-bookworm
ARG PG_DOCKER_VERSION
```

images/pgmq-pg/Dockerfile
tavallaie: Before this pull request, I was just a simple Python developer. Now, thanks to your patience and guidance, I've leveled up! Thank you!

images/pgmq-pg/Dockerfile
```diff
@@ -29,18 +31,20 @@ RUN $HOME/.cargo/bin/cargo pgrx install --pg-config=$(which pg_config)
 
 # install pg_partman
 RUN git clone https://github.com/pgpartman/pg_partman.git && \
-    cd pg_partman && \
-    git checkout v4.7.4 && make && make install
+	cd pg_partman && \
+	git checkout v4.7.4 && make && make install
 
-FROM postgres:15.4-bookworm
+ARG PG_DOCKER_VERSION=$PG_DOCKER_VERSION
```
v0idpwn, line 37:
```suggestion
ARG PG_DOCKER_VERSION=${PG_DOCKER_VERSION}
```
not sure but seems to work here (?)

images/pgmq-pg/Dockerfile
```diff
@@ -12,15 +14,15 @@ RUN apt-get update \
...
+	postgresql-server-dev-${PG_DOCKER_VERSION}
 
 # Install Rust dependencies
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 RUN $HOME/.cargo/bin/rustup default stable
 
 # init pgrx
 RUN $HOME/.cargo/bin/cargo install cargo-pgrx --version=0.11.0 --locked
-RUN $HOME/.cargo/bin/cargo pgrx init --pg15 $(which pg_config)
+RUN $HOME/.cargo/bin/cargo pgrx init --pg$${PG_DOCKER_VERSION} $(which pg_config)
```
v0idpwn, line 25:
```suggestion
RUN $HOME/.cargo/bin/cargo pgrx init --pg${PG_DOCKER_VERSION} $(which pg_config)
```

tavallaie:
should we pass it as two different ARGs? based on [moby project](https://github.com/moby/moby/issues/37345)
it should be fixed now but some how we don't see expected result.

v0idpwn:
I'm not sure. In my machine it works, but I'm using a newer version (26.1.3), the CI is running 20.10.23

images/pgmq-pg/Dockerfile
```diff
@@ -29,18 +31,20 @@ RUN $HOME/.cargo/bin/cargo pgrx install --pg-config=$(which pg_config)
 
 # install pg_partman
 RUN git clone https://github.com/pgpartman/pg_partman.git && \
-    cd pg_partman && \
-    git checkout v4.7.4 && make && make install
+	cd pg_partman && \
+	git checkout v4.7.4 && make && make install
 
-FROM postgres:15.4-bookworm
+ARG PG_DOCKER_VERSION=${PG_DOCKER_VERSION}
```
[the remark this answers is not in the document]
ChuckHend: Or was this was we tried previously?

v0idpwn:
@tavallaie I think we can hold this PR for a while. We will do a transition from pgrx to pgrxs, which will simplify our build process and allow this dockerfile to be a single stage (what would magically solve the problem)

tavallaie:
Greate, should I close it or leave it open?

v0idpwn:
Let's leave it open for now :)

v0idpwn:
Closing this in favor of #237. I went for a "Dockerfile template" which then has variables replaced. I'd be really happy to change that if we find a better approach (maybe taking 2 build args?). Still, this one will have too many conflicts after merge, as there are many CI changes and changes in this image, so probably starting clear will be better

PR #18, opened 2023-08-05
pg_partman as optional dependency

ChuckHend:
pg_partman is only required when using partitioned queues, which is not the default queue in the extension. Let's not make it a hard requirement, which will make it easier to install.

Returns to SQL caller: 
```text
SQL Error [22000]: ERROR: missing dependency: pg_partman
```

And in the postgres logs:

```text
2023-08-07 15:25:33.333 CDT [5263] WARNING:  pg_partman not installed. Install https://github.com/pgpartman/pg_partman and then run `CREATE EXTENSION pg_partman;`
2023-08-07 15:25:33.334 CDT [5263] ERROR:  missing dependency: pg_partman
2023-08-07 15:25:33.334 CDT [5263] STATEMENT:  select pgmq_create_partitioned('z')
```

Also clearing up some clippy warnings in the PR

PR #253, opened 2024-06-08
update extension contributing guide

ChuckHend:
- minor cleanups and updates
- adds some guide content around `pgenv`, which I think is useful now that we dont have `pgrx` based Postgres environments are part of the tooling
- minor Makefile additions
- removed Cargo.lock from the integration tests (there's no binary being distributed from that proejct)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,6 +81,12 @@ If you are working with partitioned queues, you will need to install `pg_partman
 make install-pg-partman
 ```
 
+Then,
+
+```sql
+CREATE EXTENSION pg_partman
+```
+
 ## Running tests
 
 Tests are written in Rust, so you will need to have the [Rust toolchain](https://www.rust-lang.org/tools/install) installed in order to run them.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,7 +84,7 @@ make install-pg-partman
 Then,
 
 ```sql
-CREATE EXTENSION pg_partman
+CREATE EXTENSION pg_partman;
 ```
 
 ## Running tests
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -100,7 +100,7 @@ PGMQ Postgres Extension releases are automated through two Github workflows; [Co
 
 ### Container Images
 
-Postgres images with PGMQ and all required dependencies are built and published to `quay.io/tembo-pg{VERSION>-pgmq:{TAG}` for all supported Postgres versions and PGMQ releases.
+Postgres images with PGMQ and all required dependencies are built and published to `quay.io/tembo-pg{PG_VERSION}-pgmq:{TAG}` for all supported Postgres versions and PGMQ releases.
 
 ### Extension Packages
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,6 +91,12 @@ Once you have a postgres instance with the extension installed, run:
 make test
 ```
 
+If you want to run the tests on a specific Postgres instance, pass the `DATABASE_URL` environment variable along with the `make` command:
+
+```bash
+make test DATABASE_URL=postgres:postgres@myhost:5432/mydb
+```
+
 ## Releases
 
 PGMQ Postgres Extension releases are automated through two Github workflows; [Containers / Trunk packages](https://github.com/tembo-io/pgmq/blob/main/.github/workflows/extension_ci.ym), and [PGXN distribution](https://github.com/tembo-io/pgmq/blob/main/.github/workflows/pgxn-release.yml). To create a release:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -93,7 +93,10 @@ make test
 
 ## Releases
 
-PGMQ Postgres Extension releases are automated through a [Github workflow](https://github.com/tembo-io/pgmq/blob/main/.github/workflows/extension_ci.yml). To create a release, create a new tag follow a valid [semver](https://semver.org/), then create a release with the same name. Auto-generate the release notes and/or add more relevant details as needed.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -17,33 +17,92 @@ version 14 on ubuntu:
...
 
 ```bash
-DATABASE_URL=postgres:postgres@localhost:5432/postgres make test
+make test
 ```
 
-# Releases
+## Releases
+
+PGMQ Postgres Extension releases are automated through a [Github workflow](https://github.com/tembo-io/pgmq/blob/main/.github/workflows/extension_ci.yml). To create a release, create a new tag follow a valid [semver](https://semver.org/), then create a release with the same name. Auto-generate the release notes and/or add more relevant details as needed.
```
theory, line 96: Also [`pgxn-release.yml`](https://github.com/tembo-io/pgmq/blob/main/.github/workflows/pgxn-release.yml).

theory (approved):

CONTRIBUTING.md
ChuckHend: Ah yes. I'll update it describing the two workflows and what they do.

Also I will add a snip about how to bump the extension version.

CONTRIBUTING.md
```diff
@@ -17,33 +17,92 @@ version 14 on ubuntu:
...
+
+### Container Images
+
+Postgres images with PGMQ and all required dependencies are built and published to `quay.io/tembo-pg{VERSION>-pgmq:{TAG}` for all supported Postgres versions and PGMQ releases.
```
v0idpwn, line 100:
```suggestion
Postgres images with PGMQ and all required dependencies are built and published to `quay.io/tembo-pg{PG_VERSION}-pgmq:{TAG}` for all supported Postgres versions and PGMQ releases.
```

CONTRIBUTING.md
```diff
@@ -17,33 +17,92 @@ version 14 on ubuntu:
...
+
 ## Running tests
+
+Tests are written in Rust, so you will need to have the [Rust toolchain](https://www.rust-lang.org/tools/install) installed in order to run them.
+
 Once you have a postgres instance with the extension installed, run:
 
 ```bash
-DATABASE_URL=postgres:postgres@localhost:5432/postgres make test
+make test
```
v0idpwn, line 91: It can be hard to discover this environment variable, I think it's good that it's documented here

CONTRIBUTING.md
```diff
@@ -17,33 +17,92 @@ version 14 on ubuntu:
...
 ```
 
+### Installing pg_partman (optional)
+
+If you are working with partitioned queues, you will need to install `pg_partman`.
+
+```bash
+make install-pg-partman
+```
+
```
v0idpwn, line 83: I think this step should be before the "Install PGMQ to Postgres" one, as `cascade` will attempt to install both extensions

CONTRIBUTING.md
```diff
@@ -17,33 +17,92 @@ version 14 on ubuntu:
...
+pgenv use 16.3
+```
+
+Connect to Postgres:
+
+```bash
+psql -U postgres
+```
+
+Fresh install will have not have PGMQ installed.
```
v0idpwn, line 41:
```suggestion
A fresh install will have not have PGMQ installed.
```

Makefile
```diff
@@ -6,6 +6,8 @@ PG_CONFIG   ?= pg_config
 PGXS := $(shell $(PG_CONFIG) --pgxs)
 include $(PGXS)
 
+DATABASE_URL:=postgres:postgres@localhost:5432/postgres
```
v0idpwn, line 9:
```suggestion
```
We want to allow `make test` to use any database url, not a fixed database

v0idpwn: There's a default in the rust code. We may write a default here but I think we should allow it to be overriden

v0idpwn: This line is why extension upgrade tests fail, they are relying on a pgrx managed postgres and thus use a different `DATABASE_URL`

CONTRIBUTING.md
ChuckHend: `CASCADE` no longer installs pg_partman since it is no longer declared as a hard dependency in `pgmq.control`.

Makefile
ChuckHend: I updated the Makefile. Take a look, we should be able to set a default AND allow it to be overridden.

CONTRIBUTING.md
ChuckHend: pg_partman was dropped as a hard dependency in https://github.com/tembo-io/pgmq/pull/18

CONTRIBUTING.md
ChuckHend: Good idea. I added a section covering this.

Makefile
ChuckHend: Ah I see, looks like we're setting it to [DATABASE_URL=postgres://$USER:postgres@localhost:28816/pgmq](https://github.com/tembo-io/pgmq/blob/c01dd3450d0a51ee5dd45b840db361ddcbc3b7b9/.github/workflows/extension_upgrade.yml#L92). This should work again.

CONTRIBUTING.md
ChuckHend:
```
postgres=# \dx
                 List of installed extensions
  Name   | Version |   Schema   |         Description          
---------+---------+------------+------------------------------
 plpgsql | 1.0     | pg_catalog | PL/pgSQL procedural language
(1 row)

postgres=# create extension pgmq cascade;
CREATE EXTENSION
postgres=# \dx
                                     List of installed extensions
  Name   | Version |   Schema   |                             Description                             
---------+---------+------------+---------------------------------------------------------------------
 pgmq    | 1.3.3   | pgmq       | A lightweight message queue. Like AWS SQS and RSMQ but on Postgres.
 plpgsql | 1.0     | pg_catalog | PL/pgSQL procedural language
(2 rows)
```

Makefile
ChuckHend: I see now. Can't go back in time and change the Makefile, so I'll just remove this part completely.

CONTRIBUTING.md
v0idpwn: Ohh, I see! But then you should ask them to CREATE EXTENSION pg_partman too, right?

v0idpwn: Ah, I see you added

v0idpwn (approved):

PR #259, opened 2024-06-12
fix LICENSE in README

tavallaie:


ChuckHend (approved):

ChuckHend (approved):

ChuckHend:
We will need to update license in Cargo.toml, and bump patch version too so that we update the license shown at crates.io

tavallaie:
Cargo.toml license updated but I  don't change version so you can change it whenever it ready for publish.

ChuckHend:
We can release it now. I went ahead and bumped that.

PR #261, opened 2024-06-13
Port (most of) the Rust tests to pg_regress

theory:
Pretty much a direct port from
`integration_test/tests/integration_tests.rs` to `test/sql/base.sql`, with the expected output in `test/expected/base.out`. The only tests not ported here are `test_transaction_send()` and `test_transaction_read()`, because they require multiple connections. Will have to look into TAP tests for that.

Update the `Makefile` with the configuration necessary for `make installcheck` to run the tests, and also simplify the line reading the version from `pgmq.control`. Tell Git to ignore temporary files output by `installcheck`, as well as `.vscode` directory (to which I add spellings).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ sql/pgmq--*.sql
 images/pgmq-pg/Dockerfile
 META.json
 Trunk.toml
+.vscode/
+/results/
+/regression.*
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,12 @@
 EXTENSION    = pgmq
-EXTVERSION   = $(shell grep "^default_version" pgmq.control | sed -E "s/^default_version[[:space:]]*=[[:space:]]*'(.*)'/\1/" | tr -d "'")
+EXTVERSION   = $(shell grep "^default_version" pgmq.control | sed -r "s/default_version[^']+'([^']+).*/\1/")
+DATA         = $(wildcard sql/*--*.sql)
+TESTS        = $(wildcard test/sql/*.sql)
+REGRESS      = $(patsubst test/sql/%.sql,%,$(TESTS))
+REGRESS_OPTS = --inputdir=test
 
-DATA 		     = $(wildcard sql/*--*.sql)
 PG_CONFIG   ?= pg_config
+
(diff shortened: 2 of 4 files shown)


theory:
Not sure what's up with the failing test. @ChuckHend has the location of `pg_config` changed?

ChuckHend:
> has the location of `pg_config` changed?

No I dont think so, but we did start setting it in the Makefile, https://github.com/tembo-io/pgmq/blob/895884c9d84f69996300aa3bec7c19af08c19af2/Makefile#L5

I wonder if that is conflicting with what we are doing in the upgrade workflow here https://github.com/tembo-io/pgmq/blob/895884c9d84f69996300aa3bec7c19af08c19af2/.github/workflows/extension_upgrade.yml#L57https://github.com/tembo-io/pgmq/blob/895884c9d84f69996300aa3bec7c19af08c19af2/Makefile#L5

v0idpwn:
I'm not sure why it didn't happen before (maybe cache?) but this seems related to #237 
As you can see here: https://github.com/tembo-io/pgmq/actions/runs/9506460473/job/26204647381?pr=261#step:6:45 , the pgx-install step doesn''t install pgrx because it finds on Cargo.toml that it's not needed. Then later it fails because we expect a pgrx-managed postgres instance to be available for this test.

ChuckHend:
@theory , it looks like the pgrx-init action [skips the initialization](https://github.com/tembo-io/pgmq/blob/895884c9d84f69996300aa3bec7c19af08c19af2/.github/actions/pgx-init/action.yml#L35) if it doesn't find pgrx in Cargo.toml


A snip from the log in the .[/.github/actions/pgx-init step](https://github.com/tembo-io/pgmq/actions/runs/9506460473/job/26204647381?pr=261)

```
  env:
    CACHE_ON_FAILURE: false
    CARGO_INCREMENTAL: 0
pgrx is not a dependency: skipping
```

ChuckHend:
We could either edit the action, or rip the action out completely and set up postgres another way 🤔

theory:
Would it work to move `pgx-init` _after_ the `Checkout old version (1.0.0)` step?

ChuckHend (approved):

PR #262, opened 2024-06-13
Port test_transaction_send & test_transaction_read

theory:
Use ISOLATION tests to manage multiple database connections. The port is pretty straightforward, although `pg_isolation_regress` is only installed on Postgres 14 and higher. So don't the `ISOLATION` variables in the `Makefile` for earlier versions.

Also use the `pgxn` client so simplify the installation of `pg_partman` in `extension_ci.yml`, and remove the Rust configuration for running tests, as they are no longer used.

Also: use `EXTRA_CLEAN` to add additional files to the `clean` target, rather than override the target defined by PGXS.

### DIFF
--- a/.github/workflows/extension_ci.yml
+++ b/.github/workflows/extension_ci.yml
@@ -55,23 +55,7 @@ jobs:
     steps:
       - run: pg-start ${{ matrix.pg }}
       - uses: actions/checkout@v4
-      - run: apt install -y pkg-config
-      - name: Install Rust stable toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          toolchain: stable
-      - uses: Swatinem/rust-cache@v2
-        with:
-          prefix-key: "pgmq-extension-test-${{ matrix.pg }}"
-          workspaces: |
-            pgmq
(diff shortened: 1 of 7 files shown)


ChuckHend (approved):

v0idpwn (approved):

.github/workflows/extension_ci.yml
```diff
@@ -55,23 +55,7 @@ jobs:
...
-          workspaces: |
-            pgmq
-      - name: Install pg_partman
-        run: |
-          git clone https://github.com/pgpartman/pg_partman.git && \
-          cd pg_partman && \
-          git checkout v4.7.4 && \
-          make && \
-          make install
+      - run: pgxn install pg_partman
```
ChuckHend, line 58: The pg_partman repo is on 5.x but I see pgxn is still on 4. I think this should probably pin to 4.7.3 until we update the project to work with pg_partman 5. And better yet, put the `Makefile` on `pgxn install pg_partman --version 4.7.4`

theory: Ah, okay. Done in 2aeca95.

Issue #257, opened 2024-06-12
Transaction Support (Rust)


haydenflinner:
# Background
One of the reasons for choosing to use postgres as your message queue rather than a second system as queue is that you don't have to worry about idempotence.
That is, if all of the state for your application is in this one Postgres instance, you can do something like
```rust
let transaction = conn.begin_transaction()
let msg = pgmq.read_and_remove(transaction)
modify_pg_bizlogic_tables(msg, transaction)
transaction.end()
```
And if you fail at any point, the transaction gets rolled back, and you're all good. This is a strictly safer approach than the one presented by having a separate msg queue, which is that you modify your biz logic tables, and _then_ you `commit` the msg read in the queue (in pgmq, this would be `pgmq.archive`), and you cross your fingers that either (1) you don't crash between doing the side effects and committing the read or (2) your side effects are idempotent, it won't matter if you read the same msg again.

# Solution
If pgmq supported receiving an [ongoing transaction ](https://stackoverflow.com/a/77406271) rather than managing its own connection, this would be trivial, I could just begin the transaction, pass it in to pgmq and otherwise use it as normal, and commit the transaction when I'm done.

ChuckHend:
Thanks @haydenflinner, just to clarify you are saying this is a limitation with the Rust client library, not the extension itself, correct? You should be able to get the desired behavior in Rust by executing SQL directly.

haydenflinner:
Right; this could be done in plain SQL, it's just the Rust client library doesn't support this usage with the convenient API.

I may give this little refactor a shot today, found another [example](https://github.com/launchbadge/sqlx/blob/main/examples/postgres/transaction/src/main.rs). Biggest hurdle will be me figuring out how to vendor the code for my own usage 😄 I guess I could fork this repo and give Cargo the URL of my fork, but that would be a pretty slow feedback loop!

v0idpwn:
I agree pgmq-rs should support easily wrapping statements in a transaction. In the elixir library, one can simply:

```elixir
MyApp.Repo.transaction(fn ->
  MyApp.Pgmq.send("q1", message)
  MyApp.Pgmq.send("q2", message)
end)
```

Seems hard to achieve something similar in pgmq-rs without breaking compatibility significantly :/

v0idpwn:
I'd love to see the same in the python client, cc @ChuckHend @tavallaie

ChuckHend:
> IMO this is one of the big selling points of pgmq over queues outside the database

[...]

PR #268, opened 2024-06-15
pgmq-python: adding support for Transaction

tavallaie:
due to #257, I try to add transaction to  python library.


tavallaie:
I don't know why my tests will fails, logger don't show anything particular. @v0idpwn @ChuckHend can help me to solve it?

v0idpwn:
I can take a look tomorrow.

tembo-pgmq-python/tembo_pgmq_python/queue.py
```diff
@@ -25,6 +37,36 @@ class QueueMetrics:
...
+    def wrapper(self, *args, **kwargs):
+        perform_transaction = kwargs.pop("perform_transaction", self.perform_transaction)
+        if perform_transaction:
+            with self.pool.connection() as conn:
+                try:
+                    logger.debug(f"Transaction started with conn: {conn}")
+                    with conn.transaction():
+                        result = func(self, *args, conn=conn, **kwargs)
+                    logger.debug(f"Transaction completed with conn: {conn}")
+                    return result
```
v0idpwn, line 53: I don't fully grasp this code, too. What makes it not start a transaction every single call?

tembo-pgmq-python/tests/test_integration.py
```diff
@@ -193,6 +192,55 @@ def test_validate_queue_name(self):
...
+    def test_transaction_create_queue(self):
+        """Test creating a queue within a transaction."""
+        try:
+            self.queue.create_queue("test_queue_txn", perform_transaction=True)
+            raise Exception("Intentional failure")
+        except Exception:
+            pass
+        finally:
+            queues = self.queue.list_queues(perform_transaction=False)
+            self.assertNotIn("test_queue_txn", queues)
```
ChuckHend, line 204: @tavallaie , I think during the transactional flow, we basically want users to control when transactions are opened and closed. In these test cases then I think I would expect there to be like a `.begin()` and a `.commit()` that are called before/after (out side of pgmq) the pgmq calls.

I'll need to look up the exact syntax in psycopg3, but in pseudo code, something like this:

```python
tx = connection.begin()
queue.create('somequeue', tx)
tx.execute('insert into some other table")
queue.send('somequeue', 'message', tx)
tx.commit()
```

I don't see a commit anywhere, so that might be why these tests are failing?

tavallaie: thanks, I find my bug I didn't have any `.commit()` in my code.

tavallaie:
> (1) Could you please add a test/example where the transaction is used by queries **outside** of pgmq? I don't exactly follow how this would be done. Assume that, for example, I wanted to: open a transaction, read 10 messages, **insert 10 entries in a misc table**, delete the 10 messages I've read, commit the transaction.

I think I fixed the bug, if it is still necessary I will add it.

tembo-pgmq-python/tests/test_integration.py
```diff
@@ -193,6 +193,55 @@ def test_validate_queue_name(self):
...
+            message = self.queue.read(self.test_queue, perform_transaction=False)
+            self.assertIsNotNone(message, "Message expected in queue")
+
+    def test_transaction_rollback(self):
+        """Test rollback of a transaction."""
+        try:
+            self.queue.send(
+                self.test_queue, self.test_message, perform_transaction=True
+            )
+            raise Exception("Intentional failure to trigger rollback")
```
ChuckHend, line 238: I think we need to raise something specific here. The way this is currently written, `self.queue.send()` can fail for any reason and the test will still pass.

ChuckHend:
@tavallaie , can you put forth an example (or better yet, a test) using this implementation with an existing transaction? This is roughly how I would imagine someone might want to use it:

```python
with psycopg.connect() as conn:
   query_results = conn.execute("select users from notifications")
   queue.tx = conn: # or something like this
   for user in query_results:
     queue.send('myq', user)
   conn.execute("delete from notifications where ....")

# a select, multiple pgmq.send(), and a delete all executed in single transaction
```

Maybe this is possible with what is currently in this PR? I am not sure though, so I think a test like this would be great. I can help with that if needed.

tavallaie:
> Maybe this is possible with what is currently in this PR? I am not sure though, so I think a test like this would be great. I can help with that if needed.

I define a decorator to do everything within single transaction but I have to write test for that decorator, I am working on it.

tavallaie:
update: I don't be able to write test for this decorator, I will share a python example tonight for how its works.

ChuckHend:
@tavallaie, are you still working on this? If not, maybe we can find someone to finish it or add transaction support in with https://github.com/tembo-io/pgmq/issues/290

tavallaie:
let me give it another shot, but I will appreciate any help I can get.

tavallaie:
let me give it another shot. but I will appreciate any help I can get.

tavallaie:
I also disagree with #290, in this way we have problem using it in Django, celery or other ORM like peewee, sqlmodel and etc. 
it is better too have different library for that.

tavallaie:
@ChuckHend  I think it is ready for review

tavallaie:
> I also disagree with #290, in this way we have problem using it in Django, celery or other ORM like peewee and etc. it is better too have different library for that.

adding some ORM natively means adding unnecessary dependency for other projects that not using those dependency at all. 
so for example If I want to make a worker without need orm or using django that already have it owns database backend we have extra dependency. 
so I suggest making another library or extra package for that like async.

ChuckHend:
@tavallaie can you add an example that demonstrates something like the SQL below, but using the python implementation in this PR?

```sql
begin;
select pgmq.create('somequeue');
insert into some_table (name) values ('brian');
select pgmq.send('somequeue', '{"hello": "world"}');
commit;
```

tavallaie:
I'll add it now

tavallaie:
I added an example directory and perform send and read all together

tembo-pgmq-python/README.md
```diff
@@ -1,242 +1,290 @@
...
-    port="5432",
-    username="postgres",
-    password="postgres",
-    database="postgres"
-)
+@transaction
+def transactional_operation(queue: PGMQueue, conn=None):
+    # Perform multiple queue operations within a transaction
+    queue.create_queue("transactional_queue", conn=conn)
+    queue.send("transactional_queue", {"message": "Hello, World!"}, conn=conn)
```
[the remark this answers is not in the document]
tavallaie: > @tavallaie , can we add some non-pgmq operations into this transaction too? for example, something similar to this?

I update example for non-pgmq and non-db success and failure for commit and rollback

tembo-pgmq-python/README.md
```diff
@@ -1,242 +1,290 @@
...
-```bash
-export PG_HOST=127.0.0.1
-export PG_PORT=5432
-export PG_USERNAME=postgres
-export PG_PASSWORD=postgres
-export PG_DATABASE=test_db
-```
-
-Initialize a connection to Postgres using environment variables:
+ # Tembo's Python Client for PGMQ
```
ChuckHend, line 1: @tavallaie - what happened with the formatting on this readme? It looks like every line got pre-pended with a space.
```suggestion
# Tembo's Python Client for PGMQ
```

ChuckHend:
@tavallaie, I think this is just about good to go, thank you for the hard work. I've approved, but please take another look at the formatting in the readme/doc (it seems like every like got indented by 1 space.

As a side note, I think many will find the decorator useful but using the psycopg connection objects directly is fairly simple too.

```python
from tembo_pgmq_python import PGMQueue
from psycopg_pool import ConnectionPool

url = "postgresql://postgres:postgres@localhost:5432/postgres"
pool = ConnectionPool(url, open=True)
queue = PGMQueue()
with pool.connection() as conn:
    with conn.transaction():
        queue.create_queue("cat2", conn=conn)
        queue.send("cat2", {"hello": "world"}, conn=conn)
        rows = conn.execute("select queue_name from pgmq.metrics_all()").fetchall()
        print(rows)
        conn.execute("create table test (id serial primary key, name text)")
        conn.execute("insert into test (name) values ('test')")
        rows = conn.execute("select * from test").fetchall()
        print(rows)
        raise Exception("This is an exception")

```

ChuckHend (approved):

PR #270, opened 2024-06-18
move test dir to pgmq-extension

ChuckHend:


v0idpwn (approved):

ChuckHend:
@v0idpwn , `make installcheck` returns status 0 when it does not find any tests. installcheck is executed by pg-build-test.

I think we'd have to patch [pg-build-test](https://github.com/pgxn/docker-pgxn-tools/blob/main/bin/pg-build-test) or maybe something trickier if we wanted that to fail.

Seems like `installcheck` should fail if no tests are run...

theory (approved):

PR #273, opened 2024-06-28
transaction support for rust client

ChuckHend:
Proposal to address https://github.com/tembo-io/pgmq/issues/257

Add an `Option<transaction>` parameter to all function calls. When the transaction is present then the function calls will use it. Otherwise they will use the connection pool created on `pgmq::PGMQueueExt::new()` or `pgmq::PGMQueueExt::new_with_pool()`

So far I've only implemented this for `.send()` to see how its going to work. If we like this then I can go ahead and implement on all the remaining function calls.


ChuckHend:
One alternative I can think of is adding a completely new object for this. I'm not sure which approach would be most idiomatic for Rust.

For example, something like this...

```rust
let pool = newpool(); 
let queue = pgmq::PGMQueueTransactional::new();
let tx  = pool.begin();
queue.send('myqueue', ... , &mut tx) # tx would be required
```

v0idpwn:
My only suggestion is that (as we discussed) I think it would be better if we received any executor instead of necessarily a transaction. 

@vrmiguel I'm curious about what you think would be the most idiomatic API for this PR

vrmiguel:
> My only suggestion is that (as we discussed) I think it would be better if we received any executor instead of necessarily a transaction.
> 
> @vrmiguel I'm curious about what you think would be the most idiomatic API for this PR

@ChuckHend and I tried out some approaches for making `cxn` generic but usability would not look good, e.g. users would have to annotate `None` as something like `None::<&mut PgConnection>` everytime

------ 

What if we had two separate functions, `send` (kept as is currently), and then `send_txn`, which takes the additional `Transaction` variable (no longer optional), to use within transactions? This way users can still mingle pgmq queries with other queries in a transaction but users just wanting to use  regular `send` don't have to keep adding an additional `None` argument every time?

ChuckHend:
>  and then send_txn, which takes the additional Transaction variable (no longer optional)

@vrmiguel, @v0idpwn  what if this were `send_with_cxn()` and it accepts any sqlx executor? So basically what @v0idpwn already [mentioned](https://github.com/tembo-io/pgmq/pull/273#issuecomment-2196919049) + what @vrmiguel suggests by creating new functions.

One thing that I think is not ideal about this is that we will nearly double the number of functions in the Rust client, since we'll want `read_with_cxn()`, `send_with_cxn()`, `archive_with_cxn()`, `delete_with_cxn()`, etc. I think thats ok though, IMO.

ChuckHend:
> Seems okay for me. I'd rather have a bit more complexity/code in our side for a better API for the dependants.

I'll get this PR updated w/ new function calls that take in a sqlx executor.

ChuckHend:
@v0idpwn , I added most of the `with_cxn()` variants of all the pgmq-rs functions. To simplify things a bit internally, I refactored all the existed functions so that they all call their corresponding `with_cxn()` variant. This makes the `*_with_cxn()` functions more like a lower level API that are used by all of the existing function calls. For example, `send()` calls `send_with_cxn()` and simply passes self.connection into it.

All `*_with_cxn()` variants are on the public API, so they will be available to all users.

vrmiguel (approved):

Issue #229, opened 2024-05-24
improve management of queue archives


ChuckHend:
pgmq.archive() deletes the message from the queue and inserts it into the queue's archive table. Messages will remain in the archive table indefinitely. Depending on volume and available storage, that can become a big problem. There's nothing preventing users from managing the queue archives themselves, but I think ideally the PGMQ API provides this for users.

In the Tembo platform, we are partitioning the archive table with pg_partman manually. Below is how we migrate an existing archive table into a partitioned table. Note: this does not migrate existing archive data to the new partitioned archive table:

Here the queue name is "data_plane_events".

```sql
BEGIN;

ALTER TABLE pgmq.a_data_plane_events RENAME TO a_data_plane_events_old;

-- DDL from existing table / same as PGMQ def but with partitioning
CREATE TABLE pgmq.a_data_plane_events (
	msg_id bigint NOT NULL,
	read_ct int4 NULL DEFAULT 0,
	enqueued_at timestamptz NULL DEFAULT now(),
	archived_at timestamptz NULL DEFAULT now(),
	vt timestamptz NULL,
	message jsonb NULL
)
PARTITION BY RANGE (archived_at);
ALTER INDEX pgmq.archived_at_idx_data_plane_events RENAME TO archived_at_idx_data_plane_events_old;
CREATE INDEX pgmqarchived_at_idx_data_plane_events ON pgmq.a_data_plane_events (archived_at);

-- initialize partman
SELECT public.create_parent(
    'pgmq.a_data_plane_events',
    'archived_at',
    'native',
    'daily'
);
-- set retention
UPDATE part_config
    SET retention = '30 days',
        retention_keep_table = false,
        retention_keep_index = false,
        infinite_time_partitions = true
    WHERE parent_table = 'pgmq.a_data_plane_events';

select public.run_maintenance();

COMMIT;
```

Once data ages out the aged partitions are just dropped and that data is lost. I think a better solution would be have the option to move the partitions to S3 instead of dropping them. We could do this by integrating with [pg_tier](https://github.com/tembo-io/pg_tier). Those aged partitions would move to S3 and remain queryable (albeit slower). This would effectively provide a "bottomless" storage for the queue archives.

ChuckHend:
partitioning added for archives in https://github.com/tembo-io/pgmq/pull/282 

I think an integration with [pg_tier](https://github.com/tembo-io/pg_tier) would still be beneficial but will create a new issue to track that.

riderx:
[...]

PR #277, opened 2024-07-11
Adding new function to convert archive queues to partitioned queues

nhudson:
This change will add a new function `pgmq.convert_archive_partitioned` which will convert an archive queue to be partitioned.

Issue: #229

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,10 +78,6 @@ CREATE EXTENSION pgmq cascade;
 If you are working with partitioned queues, you will need to install `pg_partman` version <= 4.7.0
 
 ```bash
-git clone https://github.com/pgpartman/pg_partman.git
-cd pg_partman
-git fetch
-git checkout 4.7.0
 make install-pg-partman
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,7 @@ A fresh install will have not have PGMQ installed.
 ```psql
 postgres=# \dx
                  List of installed extensions
-  Name   | Version |   Schema   |         Description          
+  Name   | Version |   Schema   |         Description
 ---------+---------+------------+------------------------------
  plpgsql | 1.0     | pg_catalog | PL/pgSQL procedural language
 (1 row)
@@ -75,9 +75,13 @@ CREATE EXTENSION pgmq cascade;
 
 ### Installing pg_partman (optional)
 
-If you are working with partitioned queues, you will need to install `pg_partman`.
+If you are working with partitioned queues, you will need to install `pg_partman` version <= 4.7.0
 
 ```bash
+git clone https://github.com/pgpartman/pg_partman.git
+cd pg_partman
+git fetch
+git checkout 4.7.0
 make install-pg-partman
 ```
 
--- a/pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
+++ b/pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
@@ -1,39 +1,67 @@
 -- Add function to convert archive queues to partitioned tables
-CREATE pgmq.convert_archive_partitioned(table_name TEXT)
+CREATE FUNCTION pgmq.convert_archive_partitioned(table_name TEXT,
+                                                 partition_interval TEXT DEFAULT '10000',
+                                                 retention_interval TEXT DEFAULT '100000',
+                                                 leading_partition INT DEFAULT 10)
 RETURNS void AS $$
+DECLARE
+a_table_name TEXT;
+a_table_name_old TEXT;
+qualified_a_table_name TEXT;
+qualified_a_table_name_old TEXT;
 BEGIN
-    EXECUTE format('ALTER TABLE pgmq.a_%I RENAME TO a_%I_old', table_name, table_name);
+  PERFORM pgmq._ensure_pg_partman_installed();
+  SELECT FORMAT('%s', 'a_'|| table_name) INTO a_table_name;
+  SELECT FORMAT('%s', 'a_'|| table_name || '_old') INTO a_table_name_old;
+  SELECT FORMAT('%s.%s', 'pgmq','a_'|| table_name) INTO qualified_a_table_name;
+  SELECT FORMAT('%s.%s', 'pgmq','a_'|| table_name || '_old') INTO qualified_a_table_name_old;
 
-    EXECUTE format('
-    CREATE TABLE pgmq.a_%I (
-        msg_id bigint NOT NULL,
-        read_ct int4 NULL DEFAULT 0,
-        enqueued_at timestamptz NULL DEFAULT now(),
-        archived_at timestamptz NULL DEFAULT now(),
-        vt timestamptz NULL,
-        message jsonb NULL
-    )
-    PARTITION BY RANGE (archived_at)', table_name);
+  PERFORM c.relkind
+    FROM pg_class c
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    WHERE c.relname = a_table_name
+    AND c.relkind = 'p';
 
-    EXECUTE format('ALTER INDEX pgmq.archived_at_idx_a_%I RENAME TO archived_at_idx_a_%I_old',
-                   replace(table_name, 'a_', ''), replace(table_name, 'a_', ''));
-    EXECUTE format('CREATE INDEX pgmqarchived_at_idx_a_%I ON pgmq.a_%I (archived_at)',
-                   replace(table_name, 'a_', ''), table_name);
+  IF FOUND THEN
+    RAISE NOTICE 'Table %s is already partitioned', a_table_name;
+    RETURN;
+  END IF;
 
-    EXECUTE format('
-    SELECT public.create_parent(
-        ''pgmq.a_%I'',
-        ''archived_at'',
-        ''native'',
-        ''daily''
-    )', table_name);
+  PERFORM c.relkind
+    FROM pg_class c
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    WHERE c.relname = a_table_name
+    AND c.relkind = 'r';
 
-    EXECUTE format('
+  IF NOT FOUND THEN
+    RAISE NOTICE 'Table %s doesnot exists', a_table_name;
+    RETURN;
+  END IF;
+
+  EXECUTE 'ALTER TABLE ' || qualified_a_table_name || ' RENAME TO ' || a_table_name_old;
+
+  EXECUTE 'CREATE TABLE '
+            || qualified_a_table_name
+            || '( msg_id BIGINT PRIMARY KEY,
+                  read_ct INT DEFAULT 0 NOT NULL,
+                  enqueued_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+                  archived_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+                  vt TIMESTAMP WITH TIME ZONE NOT NULL,
+                  message JSONB
+                ) PARTITION BY RANGE (msg_id)';
+
+    EXECUTE 'ALTER INDEX pgmq.archived_at_idx_' || table_name || ' RENAME TO archived_at_idx_' || table_name || '_old';
+    EXECUTE 'CREATE INDEX archived_at_idx_'|| table_name || ' ON ' || qualified_a_table_name ||'(archived_at)';
+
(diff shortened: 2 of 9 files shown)


pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -0,0 +1,39 @@
...
+        enqueued_at timestamptz NULL DEFAULT now(),
+        archived_at timestamptz NULL DEFAULT now(),
+        vt timestamptz NULL,
+        message jsonb NULL
+    )
+    PARTITION BY RANGE (archived_at)', table_name);
+
+    EXECUTE format('ALTER INDEX pgmq.archived_at_idx_a_%I RENAME TO archived_at_idx_a_%I_old',
+                   replace(table_name, 'a_', ''), replace(table_name, 'a_', ''));
+    EXECUTE format('CREATE INDEX pgmqarchived_at_idx_a_%I ON pgmq.a_%I (archived_at)',
```
v0idpwn, line 20: I'm worried about max identifier name size and truncation here.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -0,0 +1,39 @@
...
+                   replace(table_name, 'a_', ''), table_name);
+
+    EXECUTE format('
+    SELECT public.create_parent(
+        ''pgmq.a_%I'',
+        ''archived_at'',
+        ''native'',
+        ''daily''
+    )', table_name);
```
theory, line 29: Try using dollar quotes!

```suggestion
    EXECUTE format($_$
    SELECT public.create_parent(
        'pgmq.%I',
        'archived_at',
        'native',
        'daily'
    )$_$, 'a_' || table_name);
```

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -0,0 +1,39 @@
...
+
+    EXECUTE format('
+    UPDATE part_config
+        SET retention = ''30 days'',
+            retention_keep_table = false,
+            retention_keep_index = false,
+            infinite_time_partitions = true
+        WHERE parent_table = ''pgmq.a_%I''', table_name);
+END;
+$$ LANGUAGE plpgsql;
```
theory, line 39: Shouldn't the data be copied from the old table to the new? I suggest doing it before creating the indexes to make it nice and fast.

theory (commented):
Use of the `%I` format needs fixing. Also, CAN I HAZ TESTS?

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
ChuckHend: Maybe copying data can be an optional parameter we add later? I think there's still value in being able to convert the archive into a partitioned table while leaving any existing data on an `_old` table or something.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
ChuckHend: I think here we can just error/fail if table name is too long by providing the same name check as we do with the queue name, via `PERFORM pgmq.validate_queue_name(queue_name)`?

v0idpwn: The thing is `validate_queue_name` assumes a max queue name of 48 or something (I don't recall the exact value), which is sufficient for the biggest identifier we create with a queue name, which, if I recall correctly, is something like `a_{queue_name}_archived_at_idx`. But if you append `_old` to it, you get 4 extra characters, which then may get truncated if the queue_name is >=45

shhnwz: > The thing is `validate_queue_name` assumes a max queue name of ~48 (I don't recall the exact value), which is sufficient for the biggest identifier we create with a queue name, which, if I recall correctly, is something like `a_{queue_name}_archived_at_idx`. But if you append `_old` to it, you get 4 extra characters, which then may get truncated if the queue_name is >=45

Index pattern is 'colname_idx_{queue_name}' we don't append 'a_'  before queue_name for index name. I was wondering if we can delete the index on _old table?

pgmq-extension/sql/pgmq.sql
```diff
@@ -772,3 +772,70 @@ BEGIN
...
+a_table_name TEXT;
+a_table_name_old TEXT;
+qualified_a_table_name TEXT;
+qualified_a_table_name_old TEXT;
+BEGIN
+  PERFORM pgmq._ensure_pg_partman_installed();
+  SELECT FORMAT('%s', 'a_'|| table_name) INTO a_table_name;
+  SELECT FORMAT('%s', 'a_'|| table_name || '_old') INTO a_table_name_old;
+  SELECT FORMAT('%s.%s', 'pgmq','a_'|| table_name) INTO qualified_a_table_name;
+  SELECT FORMAT('%s.%s', 'pgmq','a_'|| table_name || '_old') INTO qualified_a_table_name_old;
```
theory, line 791:
```suggestion
  SELECT FORMAT('%s', 'a_'|| table_name) INTO a_table_name;
  SELECT FORMAT('%s', 'a_'|| table_name || '_old') INTO a_table_name_old;
  SELECT FORMAT('%s.%s', 'pgmq','a_'|| table_name) INTO qualified_a_table_name;
  SELECT FORMAT('%s.%s', 'pgmq','a_'|| table_name || '_old') INTO qualified_a_table_name_old;
```

The trick is to use `%s` for strings and `%I` for identifiers. If you're just concatenating you don't need to use format. So:

```suggestion
a_table_name TEXT := 'a_' || table_name;
a_table_name_old TEXT := 'a_'|| table_name || '_old';
qualified_a_table_name TEXT := format('%I.%I', 'pgmq', table_name);
qualified_a_table_name_old TEXT := format ('%I.%I', 'pgmq', table_name || '_old');
BEGIN
```

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -0,0 +1,39 @@
+-- Add function to convert archive queues to partitioned tables
+CREATE pgmq.convert_archive_partitioned(table_name TEXT)
+RETURNS void AS $$
+BEGIN
+    EXECUTE format('ALTER TABLE pgmq.a_%I RENAME TO a_%I_old', table_name, table_name);
```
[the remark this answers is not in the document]
shhnwz: Thanks @theory , Taken care.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -0,0 +1,39 @@
...
+
+    EXECUTE format('
+    CREATE TABLE pgmq.a_%I (
```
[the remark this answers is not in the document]
shhnwz: Thanks @theory , Done!

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
shhnwz: Thanks @theory taken care

CONTRIBUTING.md
```diff
@@ -75,9 +75,13 @@ CREATE EXTENSION pgmq cascade;
...
 
-If you are working with partitioned queues, you will need to install `pg_partman`.
+If you are working with partitioned queues, you will need to install `pg_partman` version <= 4.7.0
 
 ```bash
+git clone https://github.com/pgpartman/pg_partman.git
+cd pg_partman
+git fetch
+git checkout 4.7.0
 make install-pg-partman
```
ChuckHend, line 85: I dont think we want both the `make` command and the `git` commands here, since these `git` commands are also part of the make command:

https://github.com/tembo-io/pgmq/blob/133356e1823fbdfde4d95b37aa3726901af39061/pgmq-extension/Makefile#L40-L42

pgmq-extension/sql/pgmq.sql
ChuckHend: was there something wrong with this suggestion?

shhnwz: Sorry my bad, I didn't notice this comment.

CONTRIBUTING.md
shhnwz: Thanks @ChuckHend, Taken care

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -0,0 +1,66 @@
...
+a_table_name TEXT;
+a_table_name_old TEXT;
+qualified_a_table_name TEXT;
+qualified_a_table_name_old TEXT;
+BEGIN
+  PERFORM pgmq._ensure_pg_partman_installed();
+  SELECT FORMAT('%s', 'a_'|| table_name) INTO a_table_name;
+  SELECT FORMAT('%s', 'a_'|| table_name || '_old') INTO a_table_name_old;
+  SELECT FORMAT('%s.%s', 'pgmq','a_'|| table_name) INTO qualified_a_table_name;
+  SELECT FORMAT('%s.%s', 'pgmq','a_'|| table_name || '_old') INTO qualified_a_table_name_old;
```
theory, line 17: The trick is to use `%s` for strings and `%I` for identifiers. If you're just concatenating you don't need to use format. So:

```suggestion
a_table_name TEXT := 'a_' || table_name;
a_table_name_old TEXT := 'a_'|| table_name || '_old';
qualified_a_table_name TEXT := format('%I.%I', 'pgmq', table_name);
qualified_a_table_name_old TEXT := format ('%I.%I', 'pgmq', table_name || '_old');
BEGIN
```

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -0,0 +1,66 @@
...
+    JOIN pg_namespace n ON n.oid = c.relnamespace
+    WHERE c.relname = a_table_name
+    AND c.relkind = 'r';
+
+  IF NOT FOUND THEN
+    RAISE NOTICE 'Table %s doesnot exists', a_table_name;
+    RETURN;
+  END IF;
+
+  EXECUTE 'ALTER TABLE ' || qualified_a_table_name || ' RENAME TO ' || a_table_name_old;
```
theory, line 41: Curious why we're not using `format()` for all these executions. Would be easier to read and maintain than concatenation.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
ChuckHend, line 17: this section is different than how it is defined in `pgmq.sql`. These need to be the same.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -0,0 +1,66 @@
...
+
+  EXECUTE 'CREATE TABLE '
+            || qualified_a_table_name
+            || '( msg_id BIGINT PRIMARY KEY,
+                  read_ct INT DEFAULT 0 NOT NULL,
+                  enqueued_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+                  archived_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+                  vt TIMESTAMP WITH TIME ZONE NOT NULL,
+                  message JSONB
+                ) PARTITION BY RANGE (msg_id)';
```
theory, line 51: Dose this table have the same format as the old one? If so, you can use `LIKE`, something like:

```sql
EXECUTE format(
    'CREATE TABLE pgmq.%I LIKE %s) PARTITION BY RANGE (msg_id)',
    'a_' || table_name, 'a_'|| table_name || '_old'
);
```

theory (approved):
Can be a bit simpler, but looks like it works.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
theory: I see now that it's correct in pgmq.sql; just need to dupe it here.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
theory: Comment applies to pgmq.sql, as well.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
shhnwz: Addressed. Thanks!

PR #280, opened 2024-07-17
Archive function

shhnwz:


shhnwz:
this is taken care under https://github.com/tembo-io/pgmq/pull/277

PR #282, opened 2024-07-25
create partition will also create partitioned archive table

shhnwz:


ChuckHend:
Looks good. Can we put the upgrade migration in a `1.4.0` file? I think this is minor, since it is adding new functionality rather than a bug fix.

pgmq-extension/sql/pgmq--1.3.3--1.3.4.sql
```diff
@@ -1,3 +1,126 @@
+--Create partition will also create partitioned archive
+CREATE FUNCTION pgmq.create_partitioned(
```
ChuckHend, line 2:
```suggestion
CREATE OR REPLACE FUNCTION pgmq.create_partitioned(
```

ChuckHend (approved):
Thank you. Let's change the migration to 1.4.0 then should be good to go.

PR #287, opened 2024-08-05
Update release pgxn

ChuckHend:
Moved UPDATING.md and README.md into `./pgmq-extension`, then symlinked them onto project root. This is so that `make pgnx-zip` (and really, `git archive`) can grap these docs when publishing to pgxn.


theory (approved):
We should consider a separate README and docs for the extension from everything else. But this makes the [PGXN release](https://pgxn.org/dist/pgmq/1.4.0/) much nicer for now.

PR #289, opened 2024-08-06
doc updates: rust transaction example, readme updates

ChuckHend:
- Refactor to the the Rust client's examples so that they can be executed with `cargo run --example`.
- Add an example for using the transaction API
- Added some more detail to the pgmq extension readme

Bumping patch ver on the Rust crate so that we update readme at [crates.io](https://crates.io/crates/pgmq)


pgmq-extension/README.md
```diff
@@ -41,24 +41,25 @@ Postgres 12-16.
...
-docker run -d --name postgres -e POSTGRES_PASSWORD=postgres -p 5432:5432 quay.io/tembo/pg16-pgmq:latest
+docker run -d --name pgmq-postgres -e POSTGRES_PASSWORD=postgres -p 5432:5432 quay.io/tembo/pg16-pgmq:latest
 ```
 
 If you'd like to build from source, you can follow the instructions in [CONTRIBUTING.md](CONTRIBUTING.md).
 
 ### Updating
 
-To update PGMQ versions, follow the instructions in [UPDATING.md](UPDATING.md).
+To update PGMQ versions, follow the instructions in [UPDATING.md](pgmq-extension/UPDATING.md).
```
theory, line 62: Does UPDATING.md need to be moved into the `pgmq-extension` directory?

pgmq-extension/README.md
```diff
@@ -185,36 +211,77 @@ SELECT pgmq.archive('my_queue', 2);
 (1 row)
 ```
 
+Or archive several messages in one operation using `msg_ids` (plural) parameter:
+
+First, send a batch of messages
+
+```sql
+
+SELECT pgmq.send_batch('my_queue', ARRAY['{"foo": "bar3"}','{"foo": "bar4"}','{"foo": "bar5"}']::jsonb[]);
```
theory, line 220: Do you want to use named parameters like the other examples?

theory (approved):

pgmq-extension/README.md
ChuckHend: It moved in https://github.com/tembo-io/pgmq/pull/287

pgmq-extension/README.md
ChuckHend: Ah yes, good call.

ChuckHend: https://github.com/tembo-io/pgmq/pull/289/commits/54609153261b96b27ade54ffa0b34f915e91c334

Issue #295, opened 2024-08-14
SQL Injection


olirice:
Some of the functions exposed in `pgmq` are susceptible to SQL injection

Here's a reproducible example:
```sql
create extension pgmq;

select pgmq.create('abc');

select
   pgmq.delete(
     'abc where false;
     create table public.attack_vector(id int); -- Any SQL can be placed here
     delete from pgmq.q_abc',
     1
  );

select * from public.attack_vector; -- Demonstrate that table was created
```

and see that the table `public.attack_vector` was created.

It looks like several other functions doing SQL interpolation with `%s` are susceptible both in the SQL and over client libraries

ChuckHend:
Thanks for raising the issue @olirice. Does changing the `%s` to `%I` resolves this? Will have a PR up soon.

olirice:
It solves this problem but introduces backwards incompatibility issues 

For example
```
-- No change
select format('delete from pgmq.q_%s;', 'foo') -- delete from pgmq.q_foo
select format('delete from pgmq.q_%I;', 'foo') -- delete from pgmq.q_foo

-- Output changed
select format('delete from pgmq.q_%s;', 'FOO') -- delete from pgmq.q_FOO;
--  ^ which actually gets interpreted as delete from pgmq.q_foo

select format('delete from pgmq.q_%I;', 'FOO') -- delete from pgmq.q_"FOO";  -- which is invalid
```
so it would break anyone with special characters in a table name.

olirice:
I think to maintain backwards compatibility and plug the issue you'd want a function like

```sql
create function sanitize_table_name(name)
  returns text
  immutable
  language sql
as $$
  select format(pgmq.'%I', lower('q_' || name))
$$;

select sanitize_table_name('foo'); -- pgmq.q_foo
select sanitize_table_name('f$o'); -- pgmq."q_f$oo"
select sanitize_table_name('Foo'); -- pgmq.q_foo
select sanitize_table_name('Fo$o'); -- pgmq."q_fo$o"
```;

**Comment by ChuckHend** [comment]
Do we need the `lower()`? Differentiating with upper and lower case are valid queue and table names.

`format(pgmq.%I ....` is how we're addressing it over in https://github.com/tembo-io/pgmq/pull/296 if you can take a peek.

**Comment by olirice** [comment]
> Do we need the lower()

for backwards compatibility you would.
If someone running v1.4.0 executes

```sql
select pgmq.create('Bar')
```
that will execute
```sql
[...]

PR #296, opened 2024-08-14
use Identifier format instead of string format

ChuckHend:
Hopefully addresses https://github.com/tembo-io/pgmq/issues/295


pgmq-extension/sql/pgmq.sql
```diff
@@ -687,33 +687,33 @@ BEGIN
...
   END IF;
 
   EXECUTE FORMAT(
     $QUERY$
-    SELECT public.create_parent('pgmq.q_%s', '%s', 'native', '%s');
+    SELECT public.create_parent('pgmq.%I', '%s', 'native', '%s');
     $QUERY$,
-    queue_name, partition_col, partition_interval
+    'q_' || queue_name, partition_col, partition_interval
   );
```
theory, line 710: Arguments to a function should be passed as string literals, not identifiers, right?

```SQL
    SELECT public.create_parent('pgmq.%I', %L, 'native', %L);
```

Honestly, though, you can just call functions:

```sql
    PERFORM public.create_parent(
        'pgmq.' || quote_ident('q_' || queue_name),
        partition_col, partition_interval
    );
````

File: pgmq-extension/sql/pgmq.sql [hunk 2]
```diff
@@ -778,16 +778,16 @@ BEGIN
         retention_keep_table = false,
         retention_keep_index = true,
         automatic_maintenance = 'on'
-    WHERE parent_table = 'pgmq.a_%s';
+    WHERE parent_table = 'pgmq.&I';
```
theory, line 781:
```suggestion
    WHERE parent_table = 'pgmq.%I';
```

theory (approved):
Big improvement, left a couple of suggestions.

pgmq-extension/sql/pgmq.sql
ChuckHend: good idea!

theory (approved):
Looks good.

PR #299, opened 2024-08-15
Handle special characters in queue names

### LINKED ISSUE #293: Hyphens and other special characters in queue names cause syntax errors
Hyphens and other special characters in queue names cause a syntax error, rather than escaping the characters so the name works.
```
postgres=# SELECT pgmq.create('image-worker-queue');
ERROR:  syntax error at or near "-"
LINE 2:     CREATE TABLE IF NOT EXISTS pgmq.q_image-worker-queue (
                                                   ^
QUERY:  
    CREATE TABLE IF NOT EXISTS pgmq.q_image-worker-queue (
        msg_id BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
        read_ct INT DEFAULT 0 NOT NULL,
        enqueued_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
        vt TIMESTAMP WITH TIME ZONE NOT NULL,
        message JSONB
    )
    
CONTEXT:  PL/pgSQL function pgmq.create_non_partitioned(text) line 5 at EXECUTE
SQL statement "SELECT pgmq.create_non_partitioned(queue_name)"
PL/pgSQL function pgmq."create"(text) line 3 at PERFORM
```

ChuckHend:
Resolves #293 and addresses some follow-ups in #295

### DIFF
--- a/pgmq-extension/sql/pgmq--1.4.1--1.4.2.sql
+++ b/pgmq-extension/sql/pgmq--1.4.1--1.4.2.sql
@@ -1,0 +1,240 @@
+CREATE OR REPLACE FUNCTION pgmq.create_non_partitioned(queue_name TEXT)
+RETURNS void AS $$
+DECLARE
+  qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+  atable TEXT := pgmq.format_table_name(queue_name, 'a');
+BEGIN
+  PERFORM pgmq.validate_queue_name(queue_name);
+
+  EXECUTE FORMAT(
+    $QUERY$
+    CREATE TABLE IF NOT EXISTS pgmq.%I (
+        msg_id BIGINT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+        read_ct INT DEFAULT 0 NOT NULL,
+        enqueued_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+        vt TIMESTAMP WITH TIME ZONE NOT NULL,
+        message JSONB
+    )
+    $QUERY$,
+    qtable
+  );
+
+  EXECUTE FORMAT(
+    $QUERY$
+    CREATE TABLE IF NOT EXISTS pgmq.%I (
+      msg_id BIGINT PRIMARY KEY,
+      read_ct INT DEFAULT 0 NOT NULL,
+      enqueued_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+      archived_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+      vt TIMESTAMP WITH TIME ZONE NOT NULL,
+      message JSONB
+    );
+    $QUERY$,
+    atable
+  );
+
+  IF NOT pgmq._belongs_to_pgmq(qtable) THEN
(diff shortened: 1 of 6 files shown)


pgmq-extension/sql/pgmq.sql
```diff
@@ -842,7 +842,6 @@ DECLARE
 a_table_name TEXT := pgmq.format_table_name(table_name, 'a');
 a_table_name_old TEXT := pgmq.format_table_name(table_name, 'a') || '_old';
 qualified_a_table_name TEXT := format('pgmq.%I', a_table_name);
-qualified_a_table_name_old TEXT := format ('pgmq.%I', a_table_name_old || '_old');
```
ChuckHend, line 845: unused variable

pgmq-extension/sql/pgmq--1.4.1--1.4.2.sql
```diff
@@ -0,0 +1,240 @@
...
+      enqueued_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+      archived_at TIMESTAMP WITH TIME ZONE DEFAULT now() NOT NULL,
+      vt TIMESTAMP WITH TIME ZONE NOT NULL,
+      message JSONB
+    );
+    $QUERY$,
+    atable
+  );
+
+  IF NOT pgmq._belongs_to_pgmq(qtable) THEN
```
ChuckHend, line 36: @theory

pgmq-extension/sql/pgmq.sql
```diff
@@ -571,11 +571,11 @@ BEGIN
     atable
   );
 
-  IF NOT pgmq._belongs_to_pgmq(FORMAT('%I', qtable)) THEN
+  IF NOT pgmq._belongs_to_pgmq(qtable) THEN
```
theory, line 574: Okay, from the code:

https://github.com/tembo-io/pgmq/blob/78a766bd750b5985ac590ac384b8d294601fe544/pgmq-extension/sql/pgmq--1.1.1--1.2.0.sql#L166-L170

It looks like `_belongs_to_pgmq()` queries `pg_class`. For that case, you don't want them quoted as identifiers, but you _do_ want them to be formatted as expected. I think this is the correct approach, since we lowercase everything, but any special characters will be preserved.

IOW, if `qtable` is `q_hello there`, that's exactly how it will be read from `pg_class`, as — long as we _consistently_ lowercase entity names.

pgmq-extension/sql/pgmq.sql
```diff
@@ -734,12 +734,12 @@ BEGIN
...
       EXECUTE FORMAT('ALTER EXTENSION pgmq ADD TABLE pgmq.%I', qtable);
   END IF;
 
   PERFORM public.create_parent(
-    'pgmq.' || quote_ident(qtable),
+    FORMAT('pgmq.%s', qtable),
```
theory, line 742: Might be worth adding a comment here that the pg_partman docs say:

> `p_parent_table` - the existing parent table. MUST be schema qualified, even if in public schema.

theory (approved):

pgmq-extension/sql/pgmq.sql
ChuckHend: done!

PR #303, opened 2024-08-24
Security Issue with SQLx 0.7.2

joelawm:
There seems to have been a critical security issue with SQLx that forced all clients to update. Here is a PR of just updating the version. Let me know if I need to make anymore changes!

[Link to RUSTSEC Security Issue](https://rustsec.org/advisories/RUSTSEC-2024-0363.html)

UPDATED:
 - Version bump from 0.7.2 to 0.8.1.

I have run the tests locally and it does seem to partial pass but it keeps throwing duplicate key constraints which seems to me indicates something wrong with the test more than just updating the version because it will pass previously failed tests and just keep swapping between failed and passed. Let me know if I should dig into this further as well.

### DIFF
--- a/pgmq-rs/Cargo.toml
+++ b/pgmq-rs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.29.1"
+version = "0.29.2"
 edition = "2021"
 authors = ["Tembo.io"]
 description = "A distributed message queue for Rust applications, on Postgres."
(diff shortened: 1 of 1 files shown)


ChuckHend:
Thank you for raising and addressing this @joelawm 

>  duplicate key constraints which seems to me indicates something wrong with the test more than just updating the version

There was an issue a while back that I thought we had resolved with tests running concurrently. I think it's basically the `create if not exists` command resulting a race condition when two tests try to create a table with the same name.

ChuckHend:
The failing test is our CI system complaining that the version in `Cargo.toml` already exists in crates.io for this project. @joelawm if you can bump that version I'll merge right away, or I can merge now and get it released separately.

joelawm:
No problem! @ChuckHend I just ran using `cargo test` I looked around briefly to see if there was a more specific command but didn't see any in the documentation. I did look briefly so thats on me if its there haha.

I'll bump the version right now, give me one second!

joelawm:
Should be good now @ChuckHend

ChuckHend:
Thank you!

ChuckHend (approved):

PR #305, opened 2024-09-03
tembo_pgmq_python: Add python async interface

szobov:
Hey there,

First, thanks for your work on making PostgreSQL even more powerful!

I wanted to adapt PGMQ, but the lack of a modern async Python interface was a showstopper.
This PR brings such implementation to help the adoption of PGMQ for Python's users.

**A few notes on implementation**:

1. The database driver is [asyncpg](https://pypi.org/project/asyncpg/). Its most widely used async driver was also adopted by [SQLAlchemy](https://github.com/sqlalchemy/sqlalchemy/blob/main/pyproject.toml#L61-L64).
2. Unfortunately, Python's async and sync functions have different interfaces by definition. Therefore, I can't provide an abstract interface that both sync and async versions will implement.
3. Also, there is no async version of `__post_init__` or `__init__`. Therefore, I added an explicit `init` method for the async version. **This is the only difference in the interface**.
4. I adopted tests for the async version but didn't adopt benches for it. If it's required, I can do it.

I'm looking forward to your comments.
Thanks!

### DIFF
--- a/tembo-pgmq-python/tembo_pgmq_python/async_queue.py
+++ b/tembo-pgmq-python/tembo_pgmq_python/async_queue.py
@@ -1,0 +1,241 @@
+from dataclasses import dataclass, field
+from typing import Optional, List
+import asyncpg
+import os
+
+from orjson import dumps, loads
+
+from tembo_pgmq_python.messages import Message, QueueMetrics
+
+
+@dataclass
+class PGMQueue:
+    """Base class for interacting with a queue"""
+
+    host: str = field(default_factory=lambda: os.getenv("PG_HOST", "localhost"))
+    port: str = field(default_factory=lambda: os.getenv("PG_PORT", "5432"))
+    database: str = field(default_factory=lambda: os.getenv("PG_DATABASE", "postgres"))
+    username: str = field(default_factory=lambda: os.getenv("PG_USERNAME", "postgres"))
+    password: str = field(default_factory=lambda: os.getenv("PG_PASSWORD", "postgres"))
+    delay: int = 0
+    vt: int = 30
+    pool_size: int = 10
+    pool: asyncpg.pool.Pool = field(init=False)
+
+    def __post_init__(self) -> None:
+        self.host = self.host or "localhost"
+        self.port = self.port or "5432"
+        self.database = self.database or "postgres"
+        self.username = self.username or "postgres"
+        self.password = self.password or "postgres"
+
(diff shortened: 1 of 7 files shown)


tembo-pgmq-python/tembo_pgmq_python/async_queue.py
```diff
@@ -0,0 +1,241 @@
...
+            Message(msg_id=row[0], read_ct=row[1], enqueued_at=row[2], vt=row[3], message=loads(row[4])) for row in rows
+        ]
+
+    async def pop(self, queue: str) -> Message:
+        """Pop a message from a queue."""
+        async with self.pool.acquire() as conn:
+            rows = await conn.fetch("SELECT * FROM pgmq.pop($1);", queue)
+
+        messages = [
+            Message(msg_id=row[0], read_ct=row[1], enqueued_at=row[2], vt=row[3], message=loads(row[4])) for row in rows
```
ChuckHend, line 168: > loads(row[4])

I have not had a chance to dig into asyncpg a ton yet. Does it read the Postgres jsonb columns as a string, thus requiring us to deserialize into an object here?

tembo-pgmq-python/tembo_pgmq_python/messages.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+@dataclass
+class QueueMetrics:
+    queue_name: str
+    queue_length: int
+    newest_msg_age_sec: int
+    oldest_msg_age_sec: int
+    total_messages: int
+    scrape_time: datetime
```
ChuckHend, line 21: 👍

tembo-pgmq-python/tembo_pgmq_python/async_queue.py
szobov: It's an excellent question.
Yes, by default. You can find it [here](https://magicstack.github.io/asyncpg/current/usage.html#type-conversion) by searching "json."

It's possible, though, to set up [per-connection type conversion](https://magicstack.github.io/asyncpg/current/usage.html#example-automatic-json-conversion).
I was thinking about implementing it, but it turned out there are not many places in the code where this conversion happens.

If you think it's crucial, I can implement it.

ChuckHend: It's not crucial. It just stood out to me because we don't have to explicity convert jsonb to dict when using psycopg. Thanks for the link to docs.

ChuckHend (approved):

Issue #306, opened 2024-09-12
pg_partman assumed to be in the public schema


olirice:
Currently pg_partman is assumed to be in the `public` schema

It's hard coded in several places
https://github.com/tembo-io/pgmq/blob/18465f5f9344b9bcaca8566563d6a628f90bfa9a/pgmq-extension/sql/pgmq.sql#L750
https://github.com/tembo-io/pgmq/blob/18465f5f9344b9bcaca8566563d6a628f90bfa9a/pgmq-extension/sql/pgmq.sql#L764

which prevents `pg_partman` being referenced by `pgmq` if `pg_patman` is installed in e.g. an `extensions` schema

ChuckHend:
Hi @olirice , do you think it would be sufficient to remove the hard coding of the public schema and instead rely on search path settings? Or would you rather see this configured some other way?

olirice:
> instead rely on search path 

there's no guarantee that search path would be set appropriately for all roles so I think it'd have to be something like 

```sql
create function _get_pg_partman_schema()
  returns text
  stable
  language sql
as $$
  select
    extnamespace::regnamespace::text
  from
    pg_extension
  where
    extname = 'pg_partman';
$$;
```

and use that with a literal format anywhere its needed

```sql
perform pgmq._ensure_pg_partman_installed();
execute format('%L.create_parent(...)', pgmq.get_pg_partman_schema());
```

if you're okay with that solution I can PR it

ChuckHend:
A PR for this would be great. Thank you @olirice

PR #307, opened 2024-09-13
Support for pg_partman 5.x.x

### LINKED ISSUE #196: Update partitioned queues to pg_partman 5
Figure out what api changes need to be to [support 5](https://github.com/pgpartman/pg_partman/blob/v5.0.1/doc/pg_partman_5.0.1_upgrade.md), then implement.

Depending on how that goes, it could be a breaking change to pgmq api, especially if the migration cannot be safely automated. If breaking, we might consider maintaining dual compatibility (we can check what partman version is installed), and just deprecate the 4.7 api.

olirice:
Updates `*partitioned` functions to support pg_partman 5.x.x while maintaining 4.x.x compatibility

resolves #196 

---
If this one gets merged I'll do #306 next week

### DIFF
--- a/pgmq-extension/sql/pgmq.sql
+++ b/pgmq-extension/sql/pgmq.sql
@@ -710,22 +710,14 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
-CREATE FUNCTION pgmq._get_pg_partman_major_version() 
-RETURNS INT AS $$
-DECLARE
-  full_version TEXT;
-  major_version TEXT;
-BEGIN
-  -- Get the full version of pg_partman extension
-  SELECT extversion INTO full_version
+CREATE FUNCTION pgmq._get_pg_partman_major_version()
+RETURNS INT
+LANGUAGE SQL
+AS $$
+  SELECT split_part(extversion, '.', 1)::INT
   FROM pg_extension
-  WHERE extname = 'pg_partman';
-
-  -- Extract the major version (part before the first dot)
-  major_version := split_part(full_version, '.', 1);
-  RETURN major_version::INT;
-END;
-$$ LANGUAGE plpgsql;
+  WHERE extname = 'pg_partman'
+$$;
 
 CREATE FUNCTION pgmq.create_partitioned(
   queue_name TEXT,
--- a/pgmq-extension/sql/pgmq.sql
+++ b/pgmq-extension/sql/pgmq.sql
@@ -920,8 +920,8 @@ BEGIN
     p_control := 'msg_id',
     p_interval := partition_interval,
     p_type := case
-      when pgmq._get_pg_partman_major_version() = 5 then 'range',
-      else 'range'
+      when pgmq._get_pg_partman_major_version() = 5 then 'range'
+      else 'native'
     end
   );
 
--- a/pgmq-extension/sql/pgmq.sql
+++ b/pgmq-extension/sql/pgmq.sql
@@ -710,6 +710,23 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
+CREATE FUNCTION pgmq._get_pg_partman_major_version() 
+RETURNS INT AS $$
+DECLARE
+  full_version TEXT;
+  major_version TEXT;
+BEGIN
+  -- Get the full version of pg_partman extension
+  SELECT extversion INTO full_version
+  FROM pg_extension
+  WHERE extname = 'pg_partman';
+
+  -- Extract the major version (part before the first dot)
+  major_version := split_part(full_version, '.', 1);
+  RETURN major_version::INT;
+END;
+$$ LANGUAGE plpgsql;
+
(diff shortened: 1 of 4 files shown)


pgmq-extension/sql/pgmq.sql
```diff
@@ -710,6 +710,23 @@ BEGIN
...
+  -- Get the full version of pg_partman extension
+  SELECT extversion INTO full_version
+  FROM pg_extension
+  WHERE extname = 'pg_partman';
+
+  -- Extract the major version (part before the first dot)
+  major_version := split_part(full_version, '.', 1);
+  RETURN major_version::INT;
+END;
+$$ LANGUAGE plpgsql;
```
theory, line 728: Pretty sure this can be an inline-able SQL function.

```suggestion
CREATE FUNCTION pgmq._get_pg_partman_major_version() 
RETURNS INT LANGUAGE SQL AS $$
    SELECT split_part(extversion, '.', 1)::INT
    FROM   pg_extension
    WHERE  extname = 'pg_partman'
$$;
```

theory (approved):
Nice!

pgmq-extension/sql/pgmq.sql
olirice: good call, updated

olirice:
Heads up, CI is currently only running tests on pg_partman 4.7.3 for pgmq. You may want to add a run on pg_partman 5.1.0 too

ChuckHend:
We can ignore the image build failures in CI. We still need to figure out a way to get the organizations secrets to be passed to the workflow when it is triggered from a fork.

ChuckHend (approved):

ChuckHend:
> Heads up, CI is currently only running tests on pg_partman 4.7.3 for pgmq. You may want to add a run on pg_partman 5.1.0 too

I think the tests will need to be slightly updated for 5.0 as well. Looks like [this test](https://github.com/tembo-io/pgmq/blob/18465f5f9344b9bcaca8566563d6a628f90bfa9a/pgmq-extension/test/sql/base.sql#L193-L197) fails on pg_partman 5.1, but at first glance I think we might be just testing unrealistic values for interval based partitions.

```sql
postgres=# \dx
                                      List of installed extensions
    Name    | Version |   Schema   |                             Description                             
------------+---------+------------+---------------------------------------------------------------------
 pg_partman | 5.1.0   | public     | Extension to manage partitioned tables by time or ID
 pgmq       | 1.4.3   | pgmq       | A lightweight message queue. Like AWS SQS and RSMQ but on Postgres.
 plpgsql    | 1.0     | pg_catalog | PL/pgSQL procedural language
```

In the test, we're setting partition interval to 1, which fails.

```sql
postgres=# select pgmq.create_partitioned('pr307', '1', '2');
ERROR:  Interval for range partitioning must be greater than or equal to 2. Use LIST partitioning for single value partitions. (Values given: p_interval: 1, p_type: range)
CONTEXT: PL/pgSQL function create_parent(text,text,text,text,text,integer,text,boolean,text,text[],text,boolean,text) line 441 at RAISE
SQL statement "SELECT public.create_parent(
    p_parent_table := fq_qtable,
    p_control := partition_col,
    p_interval := partition_interval,
    p_type := case
      when pgmq._get_pg_partman_major_version() = 5 then 'range'
      else 'native'
    end
  )"
PL/pgSQL function pgmq.create_partitioned(text,text,text) line 33 at PERFORM
DETAIL: 
HINT: 
CONTEXT:  PL/pgSQL function create_parent(text,text,text,text,text,integer,text,boolean,text,text[],text,boolean,text) line 647 at RAISE
SQL statement "SELECT public.create_parent(
    p_parent_table := fq_qtable,
    p_control := partition_col,
    p_interval := partition_interval,
    p_type := case
      when pgmq._get_pg_partman_major_version() = 5 then 'range'
      else 'native'
    end
  )"
PL/pgSQL function pgmq.create_partitioned(text,text,text) line 33 at PERFORM
```

But this seems ok at a glance.

```sql
postgres=# select pgmq.create_partitioned('pr307', '2', '2');
 create_partitioned 
--------------------
 
(1 row)

select pgmq.send('pr307', '{"hello": "world"}');
 send 
------
    1
(1 row)

postgres=# \dt pgmq.q_pr307*
                    List of relations
 Schema |      Name       |       Type        |  Owner   
--------+-----------------+-------------------+----------
 pgmq   | q_pr307         | partitioned table | postgres
 pgmq   | q_pr307_default | table             | postgres
 pgmq   | q_pr307_p0      | table             | postgres
 pgmq   | q_pr307_p2      | table             | postgres
 pgmq   | q_pr307_p4      | table             | postgres
 pgmq   | q_pr307_p6      | table             | postgres
 pgmq   | q_pr307_p8      | table             | postgres
(7 rows)
```

ChuckHend:
> But this seems ok at a glance.

And updating the `partition_interval` in tests to a value >= 2 also results in all tests passing too

ChuckHend:
thanks @olirice

PR #308, opened 2024-09-13
pg_partman 5 and pg17 tests for extension ci

ChuckHend:
- Run tests on pg_partman 4.7.3 and 5.1.0
- Add tests for pg17

I will come back and update this to install pg_partman from pgxn once 5.1.0 is available in pgxn.

### DIFF
--- a/.github/workflows/extension_ci.yml
+++ b/.github/workflows/extension_ci.yml
@@ -29,7 +29,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        pg: [12, 13, 14, 15, 16]
+        pg: [12, 13, 14, 15, 16, 17]
         partman: [4.7.3, 5.1.0]
         exclude:
           # pg_partman does not support pg < 14
--- a/.github/workflows/extension_ci.yml
+++ b/.github/workflows/extension_ci.yml
@@ -30,12 +30,13 @@ jobs:
       fail-fast: false
       matrix:
         pg: [12, 13, 14, 15, 16]
+        partman: [4.7.3, 5.1.0]
     container: pgxn/pgxn-tools
     steps:
       - run: pg-start ${{ matrix.pg }}
         working-directory: ./
       - uses: actions/checkout@v4
-      - run: pgxn install 'pg_partman=4.7.3'
+      - run: pgxn install 'pg_partman=${{ matrix.partman }}'
       - run: pg-build-test
 
   publish:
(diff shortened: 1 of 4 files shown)


.github/workflows/extension_ci.yml
```diff
@@ -29,13 +29,22 @@ jobs:
...
+        partman: [4.7.3, 5.1.0]
+        exclude:
+          # pg_partman does not support pg < 14
+          - pg: 12
+            partman: 5.1.0
+          - pg: 13
+            partman: 5.1.0
     container: pgxn/pgxn-tools
     steps:
       - run: pg-start ${{ matrix.pg }}
```
theory, line 42: I suggest you add the pg_partman version to this line.

pgmq-extension/Makefile
```diff
@@ -39,7 +40,7 @@ Trunk.toml:
 install-pg-partman:
 	git clone https://github.com/pgpartman/pg_partman.git && \
 	cd pg_partman && \
-	git checkout v5.1.0 && \
+	git checkout v${PG_PARTMAN_VER} && \
```
theory, line 43: This is just so it can be successfully installed, yes? Doesn't imply that this version has to be installed, because pgmq will work with either version, correct?

theory (approved):
Looks good, could just use the pg_partman in the step name.

.github/workflows/extension_ci.yml
ChuckHend: Do you mean pass the pg_partman version in as an argument to `pg-start`?

pgmq-extension/Makefile
ChuckHend: Correct. It runs on both 4.7 and 5.1. This is just a helper function for installing pg_partman. This particular change is just so we can configure which version is installed in CI.

ChuckHend: and we wouldnt need to make this change if pg partman 5.1.0 were already in pgxn

.github/workflows/extension_ci.yml
theory: Sorry, I meant the `name` line:

``` yaml
    name: Test 🐘 ${{ matrix.pg }} with pg_partman ${{ matrix.partman }}
```

ChuckHend: ahh yes! makes sense, good call.

ChuckHend: nice 😎 

image

PR #309, opened 2024-09-13
install partman via pgxn

ChuckHend:
install pg partman using pgxn now that latest version is published https://pgxn.org/dist/pg_partman/


theory (approved):
🎉

PR #310, opened 2024-09-16
Allow pg_partman to be installed in any schema

### LINKED ISSUE #306: pg_partman assumed to be in the public schema

olirice:
Allows pg_partman to be installed in any schema vs hard-coded to `public`

#### New Functions
- pgmq._get_pg_partman_schema()

#### Updated Functions
- pgmq.drop_queue
- pgmq.create_partitioned
- pgmq.convert_archive_partitioned

Resolves #306

### DIFF
--- a/pgmq-extension/pgmq.control
+++ b/pgmq-extension/pgmq.control
@@ -1,5 +1,5 @@
 comment = 'A lightweight message queue. Like AWS SQS and RSMQ but on Postgres.'
-default_version = '1.4.3'
+default_version = '1.4.4'
 module_pathname = '$libdir/pgmq'
 schema = 'pgmq'
 relocatable = false
--- a/pgmq-extension/sql/pgmq--1.4.3--1.4.4.sql
+++ b/pgmq-extension/sql/pgmq--1.4.3--1.4.4.sql
@@ -1,4 +1,4 @@
-CREATE FUNCTION pgmq._get_pg_partman_schema()
+CREATE OR REPLACE FUNCTION pgmq._get_pg_partman_schema()
 RETURNS TEXT AS $$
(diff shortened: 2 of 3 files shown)


theory (approved):
I don't think pg_partman is relocatable, so in theory we could set `search_path` in the [CREATE FUNCTION] statements — or just get the schema name at install time and compile it into the functions. But that's probably more of a PITA than it's worth.

olirice:
agreed

Since its an optional dependency, pg_partman could be installed after pgmq. In that case install-time interpolation wouldn't be an option.

ChuckHend:
@olirice , thanks again for this work. 1.4.3 has not yet been released. Did you want this to go out in a separate patch? I'll release 1.4.3 quick if this needs to be in 1.4.4, otherwise could we put the migration sql in https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq--1.4.2--1.4.3.sql?

olirice:
No preference from me if you'd prefer to put them together. Separate PRs was only for review-ability

ChuckHend (approved):

PR #313, opened 2024-09-26
add pg17 support

ChuckHend:
Build images, update CI and readme for pg17. pgt.dev does not yet support pg17, so we will need to update again once that happens


ChuckHend:
The docker build (postgres w/ pgmq pre installed) depends on the official Postgres image, and those have not yet been published for pg17.

vrmiguel (approved):

Issue #302, opened 2024-08-23
drop queue should infer queue type


ChuckHend:
https://github.com/tembo-io/pgmq/blob/2783a6b6d61f0f9a65a0b6cfa109d88066ce7992/pgmq-extension/sql/pgmq.sql#L453 takes in a parameter designating whether the queue to drop is partitioned. When that flag is true, pgmq will handle deleting the objects/records from pg_partman as well. So if you run `pgmq.drop_queue()` on a partitioned queue without the flag, there will be some orphaned pg_partman stuff laying around.

the `pgmq.meta` table has an attribute that tracks whether table is partitioned. So on the drop_queue() call, I think we should drop the `partitioned` and replace it with a query to something like `select is_partitioned from pgmq.meta where ....`

Saturn225:
Can I get this assigned? @ChuckHend

ChuckHend:
implemented in https://github.com/tembo-io/pgmq/pull/319

PR #319, opened 2024-10-17
Make drop queue infer queue type

Neptune650:
/claim #302

This PR makes pgmq.drop_queue infer the queue type using is_partitioned and removes the partitioned argument directly.

Also removes the usage of the argument from everywhere.

### DIFF
--- a/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
+++ b/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
@@ -77,55 +77,7 @@ BEGIN
         queue_name
     ) INTO partitioned;
 
-    EXECUTE FORMAT(
-        $QUERY$
-        ALTER EXTENSION pgmq DROP TABLE pgmq.%I
-        $QUERY$,
-        qtable
-    );
-
-    EXECUTE FORMAT(
-        $QUERY$
-        ALTER EXTENSION pgmq DROP TABLE pgmq.%I
-        $QUERY$,
-        atable
-    );
-
-    EXECUTE FORMAT(
-        $QUERY$
-        DROP TABLE IF EXISTS pgmq.%I
-        $QUERY$,
-        qtable
-    );
-
-    EXECUTE FORMAT(
-        $QUERY$
-        DROP TABLE IF EXISTS pgmq.%I
-        $QUERY$,
-        atable
-    );
-
-     IF EXISTS (
-          SELECT 1
-          FROM information_schema.tables
-          WHERE table_name = 'meta' and table_schema = 'pgmq'
-     ) THEN
-        EXECUTE FORMAT(
-            $QUERY$
-            DELETE FROM pgmq.meta WHERE queue_name = %L
-            $QUERY$,
-            queue_name
-        );
-     END IF;
-
-     IF partitioned THEN
-        EXECUTE FORMAT(
-          $QUERY$
-          DELETE FROM %I.part_config where parent_table in (%L, %L)
-          $QUERY$,
-          pgmq._get_pg_partman_schema(), fq_qtable, fq_atable
-        );
-     END IF;
+    PERFORM pgmq.drop_queue(queue_name, partitioned);
 
     RETURN TRUE;
 END;
--- a/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
+++ b/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
@@ -1,3 +1,4 @@
+DROP FUNCTION IF EXISTS pgmq.drop_queue(TEXT, BOOLEAN);
 CREATE OR REPLACE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)
 RETURNS BOOLEAN AS $$
 DECLARE
--- a/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
+++ b/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
@@ -1,5 +1,6 @@
 DROP FUNCTION IF EXISTS pgmq.drop_queue(TEXT, BOOLEAN);
-CREATE OR REPLACE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)
+
+CREATE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)
 RETURNS BOOLEAN AS $$
 DECLARE
     qtable TEXT := pgmq.format_table_name(queue_name, 'q');
@@ -7,6 +8,30 @@ DECLARE
     atable TEXT := pgmq.format_table_name(queue_name, 'a');
     fq_atable TEXT := 'pgmq.' || atable;
 BEGIN
+    RAISE WARNING "drop_queue(queue_name, partitioned) is deprecated and will be removed in PGMQ v2.0. Use drop_queue(queue_name) instead.";
+
+    PERFORM pgmq.drop_queue(queue_name);
+
+    RETURN TRUE;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE FUNCTION pgmq.drop_queue(queue_name TEXT)
+RETURNS BOOLEAN AS $$
+DECLARE
+    qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+    fq_qtable TEXT := 'pgmq.' || qtable;
+    atable TEXT := pgmq.format_table_name(queue_name, 'a');
+    fq_atable TEXT := 'pgmq.' || atable;
+    partitioned BOOLEAN;
+BEGIN
+    EXECUTE FORMAT(
+        $QUERY$
+        SELECT is_partitioned FROM pgmq.meta WHERE queue_name = %L
+        $QUERY$,
+        queue_name
+    ) INTO partitioned;
+
     EXECUTE FORMAT(
         $QUERY$
         ALTER EXTENSION pgmq DROP TABLE pgmq.%I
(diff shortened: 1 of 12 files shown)


pgmq-extension/sql/pgmq.sql
```diff
@@ -460,14 +460,22 @@ RETURNS TEXT AS $$
     extname = 'pg_partman';
 $$ LANGUAGE SQL;
 
-CREATE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN DEFAULT FALSE)
+CREATE FUNCTION pgmq.drop_queue(queue_name TEXT)
```
ChuckHend, line 463: can we overload `drop_queue()` instead so that we can maintain backwards compatibility under semver?

Then we can drop `pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN DEFAULT FALSE)` when we release v2, which does not yet have a target date.

Neptune650: Just to confirm, we shouldn't remove the "partitioned" argument just yet? Just ignore it?

ChuckHend: What I mean is to add a new function with the signature `pgmq.drop_queue(queue_name TEXT)` (that will look up the queue type) and leave `pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN DEFAULT FALSE)` the way it currently is.

pgmq-extension/test/sql/base.sql
```diff
@@ -67,11 +67,11 @@ SELECT * from {PGMQ_SCHEMA}.metrics_all();
 
 -- delete all the queues
 -- delete partitioned queues
-SELECT pgmq.drop_queue(queue, true)
+SELECT pgmq.drop_queue(queue)
```
ChuckHend, line 70: Let's keep the test for `pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)` and then add a test for `pgmq.drop_queue(queue_name TEXT)`.

pgmq-extension/sql/pgmq.sql
```diff
@@ -522,6 +522,76 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
+CREATE FUNCTION pgmq.drop_queue(queue_name TEXT)
```
ChuckHend, line 525: Please also add this new function definition to a migration file - `pgmq-extension/sql/pgmq--1.44--1.5.0.sql`.

pgmq-extension/sql/pgmq.sql
```diff
@@ -460,7 +460,7 @@ RETURNS TEXT AS $$
     extname = 'pg_partman';
 $$ LANGUAGE SQL;
 
-CREATE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN DEFAULT FALSE)
+CREATE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)
```
ChuckHend, line 463: I think there may also need to be an `alter function`, or `drop function` and `create function` command added to `pgmq-extension/sql/pgmq--1.44--1.5.0.sq` since the signature is changing.

Neptune650:
@ChuckHend I have committed the changes you requested, tell me what you think

pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
```diff
@@ -0,0 +1,131 @@
+CREATE OR REPLACE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)
```
ChuckHend, line 1: I don't think we need to add `pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)` to the migration script, since this one is not changing, right? Only `pgmq.drop_queue(queue_name TEXT)` needs to be in the migration.

ChuckHend: Oh I see, I'm wrong (default value is being dropped). We probably need to use `drop function` instead of `create or replace`

https://github.com/tembo-io/pgmq/actions/runs/11413295614/job/31774992193?pr=319

```
ERROR:  cannot remove parameter defaults from existing function
HINT:  Use DROP FUNCTION drop_queue(text,boolean) first.
```

Neptune650: Should be fixed now.

pgmq-extension/sql/pgmq.sql
```diff
@@ -528,6 +528,76 @@ BEGIN
...
+    partitioned BOOLEAN;
+BEGIN
+    EXECUTE FORMAT(
+        $QUERY$
+        SELECT is_partitioned FROM pgmq.meta WHERE queue_name = %L
+        $QUERY$,
+        queue_name
+    ) INTO partitioned;
+
+    EXECUTE FORMAT(
```
ChuckHend, line 547: Everything looks looks good, just had a though though that might make reduce some code duplication -- what if instead of duplicating all the code from here and below, we just call 

```sql
pgmq.drop_queue(&I, partitioned) 
```

then remove all the code below?

Neptune650: @ChuckHend Done now.

ChuckHend (approved):

pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
```diff
@@ -0,0 +1,84 @@
+DROP FUNCTION IF EXISTS pgmq.drop_queue(TEXT, BOOLEAN);
+CREATE OR REPLACE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)
```
v0idpwn, line 2:
```suggestion
CREATE FUNCTION pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN)
```
As it was just deleted

pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
```diff
@@ -0,0 +1,84 @@
...
+RETURNS BOOLEAN AS $$
+DECLARE
+    qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+    fq_qtable TEXT := 'pgmq.' || qtable;
+    atable TEXT := pgmq.format_table_name(queue_name, 'a');
+    fq_atable TEXT := 'pgmq.' || atable;
+BEGIN
+    EXECUTE FORMAT(
```
v0idpwn, line 10:
```suggestion
BEGIN
    RAISE WARNING "drop_queue(queue, is_partitioned) is deprecated and will be removed in pgmq 2.0. Use drop_queue(queue) instead"
    
    EXECUTE FORMAT(
```

v0idpwn (approved):

pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
v0idpwn: Ah, you can't do this because you're calling this function from the other function...

ChuckHend: It was my request to call this function from the other, just to reduce code duplication. If we'd rather have the deprecation warning, I'd be ok with going back to duplicating that code.

pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
Neptune650: https://github.com/tembo-io/pgmq/blob/main/pgmq-extension%2Fsql%2Fpgmq--1.1.1--1.2.0.sql#L39
I did it this way to be consistent with other parts of the code. Should I still apply your changes?

Neptune650:
Thanks for the reviews 
I'd appreciate if the other PRs were checked too/merged.

pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
v0idpwn: What if we move the bulk of the logic to the non-deprecated function, and make the deprecated function call the non-deprecated one?

Neptune650: That wouldn't be possible because the non-deprecated function only has 1 argument, while the deprecated one has 2.

v0idpwn: Of course it is, you just need to ignore the second one

Neptune650: Well that's what I proposed initially (https://github.com/tembo-io/pgmq/pull/319#discussion_r1805130178) but it wasn't accepted

ChuckHend: dropping the parameter is different than ignoring it, right?

What I was saying in https://github.com/tembo-io/pgmq/pull/319#discussion_r1805150303 is that if we drop the function `pgmq.drop_queue(queue_name TEXT, partitioned BOOLEAN DEFAULT FALSE)` then it is a breaking change. Making it such the the `partitioned` has the same behavior whether the value is `true` or `false` I think would be ignoring it.

Neptune650: @chuckhend So, ideally the way to go is to move the code to the non deprecated function, and have the deprecated function ignore the `partitioned` argument?

Neptune650: @ChuckHend Also I was referring to this specific part of the exchange.

v0idpwn: I'll push a commit doing the change in your branch

Neptune650: @v0idpwn that works for me, thanks.

v0idpwn: @Neptune650 lol, I tried to push and it got behind somehow. Trying to fix it, sorry about the noise

v0idpwn: Done. Hopefully it's alright now :)

Neptune650: @v0idpwn Looks pretty good to me, let me know if it can be merged yet.

v0idpwn:
Thanks, @Neptune650 !

Neptune650:
@ChuckHend @v0idpwn Thanks! Can the bounty be paid off.

Issue #315, opened 2024-10-09
feature request: possible to have send() use a specific date time not a delay


RealZimboGuy:
Hi, is it possible for a new method or way to send a message when it will become available at a specific date and time. rather than trying to calculate a future date based on the duration

ie 

    select * from pgmq.send('my_queue', '{"hello": "world"}','2024-10-20 00:00:00');

ChuckHend:
Yes this is definitely possible and we'd accept a PR for it. I would need think a bit about it, but some possible ways to do it would be to overload `pgmq.send()` so that the `delay` parameter accepts either an integer or a timestamp type. OR, perhaps we add a new function `pgmq.send_at()`.

RealZimboGuy:
I can try dive into the change and see where i end up. Will be quite a bit of learning if i have to go much into the rust side of the plugin.

ChuckHend:
The postgres extension itself is all pl/pgsql now. You can add/modify the code in [pgmq.sql](https://github.com/tembo-io/pgmq/blob/e2740d0da8e663618feb19e9c6f6653de861fe14/pgmq-extension/sql/pgmq.sql#L264), then follow the install steps in the [contributing guide](https://github.com/tembo-io/pgmq/blob/main/CONTRIBUTING.md#installing-postgres) to install the new code locally as you develop it.

Saturn225:
Can I get this assigned? I will push changes in the PR by EOD

cc: @ChuckHend

Neptune650:
@ChuckHend Hey there, I don't think I was rewarded... Could you please check it.

ChuckHend:
> @ChuckHend Hey there, I don't think I was rewarded... Could you please check it.

Sorry for delay @Neptune650. Reward should be distributed now.

Neptune650:
> > @ChuckHend Hey there, I don't think I was rewarded... Could you please check it.
> 
> Sorry for delay @Neptune650. Reward should be distributed now.

@ChuckHend Thanks, I appreciate it.

PR #320, opened 2024-10-17
Create send_at and send_batch_at functions receiving a timestamp

Neptune650:
/claim #315

This pull request creates two functions - send_at and send_batch_at - which receive a timestamp and convert it to seconds from now.

This also documents both functions.

### DIFF
--- a/pgmq-extension/sql/pgmq.sql
+++ b/pgmq-extension/sql/pgmq.sql
@@ -282,6 +282,29 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
+-- send_at
+-- sends a message to a queue, with a delay as a timestamp
+CREATE FUNCTION pgmq.send_at(
+    queue_name TEXT,
+    msg JSONB,
+    delay TIMESTAMP
+) RETURNS SETOF BIGINT AS $$
+DECLARE
+    sql TEXT;
+    qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+BEGIN
+    sql := FORMAT(
+        $QUERY$
+        INSERT INTO pgmq.%I (vt, message)
+        VALUES ((clock_timestamp() + %L), $1)
+        RETURNING msg_id;
+        $QUERY$,
+        qtable, AGE(date_trunc('second', delay), date_trunc('second', CURRENT_TIMESTAMP))
+    );
+    RETURN QUERY EXECUTE sql USING msg;
+END;
+$$ LANGUAGE plpgsql;
+
 -- send_batch
 -- sends an array of list of messages to a queue, optionally with a delay
 CREATE FUNCTION pgmq.send_batch(
@@ -300,6 +323,29 @@ BEGIN
         RETURNING msg_id;
         $QUERY$,
         qtable, make_interval(secs => delay)
+    );
+    RETURN QUERY EXECUTE sql USING msgs;
+END;
+$$ LANGUAGE plpgsql;
+
+-- send_batch_at
+-- sends an array of list of messages to a queue, with a delay as a timestamp
+CREATE FUNCTION pgmq.send_batch_at(
+    queue_name TEXT,
+    msgs JSONB[],
+    delay TIMESTAMP
+) RETURNS SETOF BIGINT AS $$
+DECLARE
+    sql TEXT;
+    qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+BEGIN
+    sql := FORMAT(
+        $QUERY$
+        INSERT INTO pgmq.%I (vt, message)
+        SELECT clock_timestamp() + %L, unnest($1)
+        RETURNING msg_id;
+        $QUERY$,
+        qtable, AGE(date_trunc('second', delay), date_trunc('second', CURRENT_TIMESTAMP))
     );
     RETURN QUERY EXECUTE sql USING msgs;
 END;
--- a/tembo-pgmq-python/tembo_pgmq_python/async_queue.py
+++ b/tembo-pgmq-python/tembo_pgmq_python/async_queue.py
@@ -176,6 +176,27 @@ class PGMQueue:
         return result[0]
 
     @transaction
+    async def send_at(self, queue: str, message: dict, delay: str, conn=None) -> int:
+        """Send a message to a queue with timestamp."""
+        self.logger.debug(f"send_at called with queue='{queue}', message={message}, delay={delay}, conn={conn}")
+        if conn is None:
+            async with self.pool.acquire() as conn:
+                return await self._send_at_internal(queue, message, delay, conn)
+        else:
+            return await self._send_at_internal(queue, message, delay, conn)
+
+    async def _send_at_internal(self, queue, message, delay, conn):
+        self.logger.debug(f"Sending message to queue '{queue}' with delay={delay}")
+        result = await conn.fetchrow(
+            "SELECT * FROM pgmq.send_at($1, $2::jsonb, $3);",
+            queue,
+            dumps(message).decode("utf-8"),
+            delay,
+        )
+        self.logger.debug(f"Message sent with msg_id={result[0]}")
+        return result[0]
+
+    @transaction
     async def send_batch(self, queue: str, messages: List[dict], delay: int = 0, conn=None) -> List[int]:
         """Send a batch of messages to a queue."""
         self.logger.debug(f"send_batch called with queue='{queue}', messages={messages}, delay={delay}, conn={conn}")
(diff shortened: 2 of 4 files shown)


pgmq-extension/sql/pgmq.sql
```diff
@@ -282,6 +282,29 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
+-- send_at
+-- sends a message to a queue, with a delay as a timestamp
+CREATE FUNCTION pgmq.send_at(
+    queue_name TEXT,
+    msg JSONB,
+    delay TIMESTAMP
+) RETURNS SETOF BIGINT AS $$
```
ChuckHend, line 291: Will it work if we overload the functions instead of defining new function names?

```suggestion
CREATE FUNCTION pgmq.send(
    queue_name TEXT,
    msg JSONB,
    vt TIMESTAMPZ
) RETURNS SETOF BIGINT AS $$
```

tembo-pgmq-python/tembo_pgmq_python/async_queue.py
```diff
@@ -198,6 +219,29 @@ async def _send_batch_internal(self, queue, messages, delay, conn):
...
+        jsonb_array = [dumps(message).decode("utf-8") for message in messages]
+        result = await conn.fetch(
+            "SELECT * FROM pgmq.send_batch_at($1, $2::jsonb[], $3);",
+            queue,
+            jsonb_array,
+            delay,
+        )
+        msg_ids = [message[0] for message in result]
+        self.logger.debug(f"Batch messages sent with msg_ids={msg_ids}")
+        return msg_ids
```
ChuckHend, line 243: @tavallaie - what are you opinion on function naming in the python client? I'm more in favor of overloading the function names in Postgres, but in the python client I am not sure.

tavallaie: In python, function overloading isn't supported. It's easier to maintain code with separate, descriptive function names. so those names are good in my opinion.

pgmq-extension/sql/pgmq.sql
Neptune650: I figured it would make more sense to be consistent between sql and python since python doesn't have function overloading

tembo-pgmq-python/tembo_pgmq_python/async_queue.py
ChuckHend: While overloading in python is not supported, parameters can accept multiple types. For example, `send(... delay: int, vt: Option[Datetime] = None)` is valid, right?

I'm ok just going with `send_batch()` in the Python client though.

ChuckHend: or even something this would be valid:

```python
async def _send_batch_at_internal(self, queue, messages, vt: Union[int, Datetime], conn):
   if isinstance(vt, int) ....
   else ...
```

pgmq-extension/sql/pgmq.sql
ChuckHend: The extension is not aware of which clients exist, so I don't want to make assumptions in the extensions that depend on clients.

However, we'd not really overloading much elsewhere so let's go ahead and keep this `send_batch()`. I will likely put together a proposal to overload everything for the pgmq v2 release though.

Neptune650:
@ChuckHend So is there anything else left to work on or can this be merged now?

ChuckHend:
> @ChuckHend So is there anything else left to work on or can this be merged now?

Test cases would be great1

ChuckHend:
You are correct, we don't have a test to ensure that the `delay` feature continues to work as intended, but there are [tests](https://github.com/tembo-io/pgmq/blob/e2740d0da8e663618feb19e9c6f6653de861fe14/pgmq-extension/test/sql/base.sql#L91) calling the function (in the future we should add tests to ensure `delay` continues to work as intended on `send()` too). Please add test cases for the new functions that are introduced in this PR.

Neptune650:
@ChuckHend I have added tests, tell me your thoughts.

v0idpwn (commented):
I think it would probably be a better API if instead `send` and `send_batch` were overloaded to support both a delay and a timestamp.

Neptune650:
> I think it would probably be a better API if instead `send` and `send_batch` were overloaded to support both a delay and a timestamp. 

@v0idpwn I've already talked with ChuckHend about this

And an issue remains and that is, which function to run when you only 2 arguments

v0idpwn:
Sorry, missed it because it was resolved :)

tembo-pgmq-python/tembo_pgmq_python/async_queue.py
tavallaie: is it possible if someone use both together? if not we can add some exception to prevent that.

tavallaie:
if both variable are not none, you can simple raise an exception to user that those variable can not be use together.

tavallaie:
another question if user add delay can change it to vt or vise versa? what will happened? what we except? 
what scenario use delay or vt or both?

Neptune650:
> if both variable are not none, you can simple raise an exception to user that those variable can not be use together.

No, the issue is that Postgres will raise an exception if there are two overloading functions with the only difference being a default parameter.

Neptune650:
> another question if user add delay can change it to vt or vise versa? what will happened? what we except? what scenario use delay or vt or both?

What do you mean?
The user can use delays and change it to vt yes, that's always been an option. The other way around can be done too with CURRENT_TIMESTAMP. It's up to the user to use timestamps or delays.

Neptune650:
Is there anything else I'm missing or is this PR done now?

tavallaie:
> > if both variable are not none, you can simple raise an exception to user that those variable can not be use together.
> 
> No, the issue is that Postgres will raise an exception if there are two overloading functions with the only difference being a default parameter.

Sorry, but I can't understand why we can't use delay as an option beside vt with default none or zero?

Neptune650:
> > > if both variable are not none, you can simple raise an exception to user that those variable can not be use together.
> > 
> > No, the issue is that Postgres will raise an exception if there are two overloading functions with the only difference being a default parameter.
> 
> Sorry, but I can't understand why we can't use delay as an option beside vt with default none or zero?  
> 

Look, if we overload two functions like this
```sql
pgmq.send_batch( 
     queue_name text, 
     msgs jsonb[], 
     delay timestamp DEFAULT CURRENT_TIMESTAMP
 );

pgmq.send_batch( 
     queue_name text, 
     msgs jsonb[], 
     delay integer DEFAULT 0
 );
```
Postgres will raise an exception because it won't know which one to run if you just put the first two parameters.

Neptune650:
@ChuckHend Fixed tests, tell me if there's anything missing now.

v0idpwn:
@Neptune650 jfyi, I plan to take a look on this PR and do some tests with it by, at most, the weekend.  Consider it accepted, but I may push some changes.

Neptune650:
> @Neptune650 jfyi, I plan to take a look on this PR and do some tests with it by, at most, the weekend.  Consider it accepted, but I may push some changes.

Sounds good to me, thanks.

Neptune650:
@v0idpwn any update on this?

v0idpwn:
@ChuckHend @tavallaie not sure what should be done with the python client after these changes :)

v0idpwn:
@Neptune650 I pushed the following changes:
- Overload `pgmq.send`/`pgmq.send_batch` instead of adding new `send_at` and `send_batch_at`
- Refactor to remove code duplication
- Remove unnecessary timestamp -> duration -> timestamp conversion

I think the extension code is fine now. We just need to figure what to do in the python api.

Neptune650:
> @Neptune650 I pushed the following changes:
> - Overload `pgmq.send`/`pgmq.send_batch` instead of adding new `send_at` and `send_batch_at`
> - Refactor to remove code duplication
> - Remove unnecessary timestamp -> duration -> timestamp conversion
> 
> I think the extension code is fine now. We just need to figure what to do in the python api.

@Neptune650 All sounds good to me, I was thinking we could leave the Python function names as is, while internally calling the overloaded pgmq.send

tavallaie:
> @Neptune650 I pushed the following changes:
> - Overload `pgmq.send`/`pgmq.send_batch` instead of adding new `send_at` and `send_batch_at`
> - Refactor to remove code duplication
> - Remove unnecessary timestamp -> duration -> timestamp conversion
> - Take timestamp with tz instead of timestamp without tz
> 
> I think the extension code is fine now. We just need to figure out what to do in the python API.

I think we can add parameters of delay and ts with a default value of none; check if those, too are not none and use overloaded functions instead.

Neptune650:
@tavallaie I have pushed the changes you suggested, please tell me what you think

tavallaie:
@Neptune650 did you update unit tests too?

tavallaie:
@Neptune650 did you update unit tests?

tembo-pgmq-python/tembo_pgmq_python/queue.py
```diff
@@ -112,20 +112,38 @@ def list_queues(self, conn=None) -> List[str]:
...
 
     @transaction
-    def send_batch(self, queue: str, messages: List[dict], delay: int = 0, conn=None) -> List[int]:
+    def send_batch(self, queue: str, messages: List[dict], delay: int = 0, tz: str = None, conn=None) -> List[int]:
         """Send a batch of messages to a queue."""
         self.logger.debug(f"send_batch called with conn: {conn}")
-        query = "select * from pgmq.send_batch(%s, %s, %s);"
-        params = [queue, [Jsonb(message) for message in messages], delay]
-        result = self._execute_query_with_result(query, params, conn=conn)
+        result = None
```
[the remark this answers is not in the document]
Neptune650: Yes, delay will be prioritized first.

Neptune650:
> @Neptune650 did you update unit tests too? 

@tavallaie There's no need, they're working fine without changes.

ChuckHend:
@Neptune650 , @tavallaie thoughts on moving the python changes to a new PR and focus on getting the SQL API changes approved? Also, while the python client changes are very helpful IMO they are out of scope to the bounty that @Neptune650 is working on.

v0idpwn:
@Neptune650 would you please revert the Python changes, then? Then I can merge.

Neptune650:
@v0idpwn I reverted the changes, please merge.

tavallaie:
> > @Neptune650 , @tavallaie thoughts on moving the python changes to a new PR and focus on getting the SQL API changes approved? Also, while the python client changes are very helpful IMO they are out of scope to the bounty that @Neptune650 is working on.
> 
> @ChuckHend Sounds good to me, I can create another PR for the Python changes (I'd appreciate if it was with a separate tip but I'll send it either way)

If you are busy, I can do that.

Neptune650:
> > > @Neptune650 , @tavallaie thoughts on moving the python changes to a new PR and focus on getting the SQL API changes approved? Also, while the python client changes are very helpful IMO they are out of scope to the bounty that @Neptune650 is working on.
> > 
> > 
> > @ChuckHend Sounds good to me, I can create another PR for the Python changes (I'd appreciate if it was with a separate tip but I'll send it either way)
> 
> If you are busy, I can do that.

@tavallaie I've created a PR for it #333

ChuckHend (approved):
Thank you @Neptune650

Neptune650:
> Thank you @Neptune650 

@ChuckHend No problem - please merge if ready.

v0idpwn (approved):

PR #321, opened 2024-10-18
added truncate instead of delete

### LINKED ISSUE #298: pgmq.purge(queue_name text) performance
In `pgmq.purge(queue_name text)` it uses `delete from` with no `where` clause and then collect the row count with `GET DIAGNOSTICS`

https://github.com/tembo-io/pgmq/blob/78a766bd750b5985ac590ac384b8d294601fe544/pgmq-extension/sql/pgmq--1.4.0--1.4.1.sql#L310-L311

If the queue is large that will create a lot of IO and WAL with 1 record per deleted tuple. At scale that could contribute to exhausting IO on the primary and cause replication lag on any replicas

What do you think about swapping it out for something like
```
SELECT count(*) INTO deleted_count FROM pgmq.my_tab;
TRUNCATE pgmq.my_tab;
```
since truncate produces a single WAL record a little IO on the primary?

Abiji-2020:
closes #298 
/claim #298 

# Modifed the code in sql to use truncate instead of delete when using purge

### DIFF
--- a/pgmq-extension/sql/pgmq.sql
+++ b/pgmq-extension/sql/pgmq.sql
@@ -378,26 +378,21 @@ END
 $$ LANGUAGE plpgsql;
 
 -- purge queue, deleting all entries in it.
-CREATE OR REPLACE FUNCTION pgmq.purge(
-    queue_name text
-)
-RETURNS integer AS $$
+CREATE OR REPLACE FUNCTION pgmq."purge_queue"(queue_name TEXT)
+RETURNS BIGINT AS $$
 DECLARE
-    qtable text;
-    deleted_count integer;
+  deleted_count INTEGER;
+  qtable TEXT := pgmq.format_table_name(queue_name, 'q');
 BEGIN
-    -- Determine the name of the queue table
-    SELECT INTO qtable pgmq.get_queue_table_name(queue_name);
+  -- Get the row count before truncating
+  EXECUTE format('SELECT count(*) FROM pgmq.%I', qtable) INTO deleted_count;
 
-    -- Get the row count before truncating
-    EXECUTE format('SELECT count(*) FROM pgmq.%I', qtable) INTO deleted_count;
-    
-    -- Use TRUNCATE for better performance on large tables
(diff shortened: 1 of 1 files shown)


pgmq-extension/sql/pgmq.sql
```diff
@@ -378,18 +378,24 @@ END
...
+CREATE OR REPLACE FUNCTION pgmq."purge_queue"(queue_name TEXT)
 RETURNS BIGINT AS $$
 DECLARE
   deleted_count INTEGER;
   qtable TEXT := pgmq.format_table_name(queue_name, 'q');
 BEGIN
-  EXECUTE format('DELETE FROM pgmq.%I', qtable);
-  GET DIAGNOSTICS deleted_count = ROW_COUNT;
+  -- Get the row count before truncating
+  EXECUTE format('SELECT count(*) FROM pgmq.%I', qtable) INTO deleted_count;
```
v0idpwn, line 388:
```suggestion
  EXECUTE format('SELECT count(*) FROM pgmq.%I', qtable) INTO deleted_count;
```

Considering the goal of this function is performance, I think it's fair to change our description to "estimation of number of rows deleted", and use the reltuples stat. I don't think the exact number of messages deleted is too important anyway.

v0idpwn: Just an idea, though.

Abiji-2020: then  shall i modify it ??

ChuckHend: Let's leave it for now since current behavior tells user how many are deleted. We can consider dropping it for the v2.0 release of the project though.

Abiji-2020:
@ChuckHend build and push images is failing on to installation of stoml, what can we do about it

ChuckHend:
Don't worry about these image build failures.

Abiji-2020:
> Don't worry about these image build failures.

Okay, waiting for review and merge 😊

ChuckHend (approved):

PR #322, opened 2024-10-19
Add new conditional parameter for read

### LINKED ISSUE #288: Feature Request: Conditional reading for PGMQ
I request an enhancement for the PGMQ extension to support conditional message retrieval. This would allow pulling based on specific attributes directly within the pgmq_read function.

eg: Pull messages with a specific type:
`SELECT * FROM pgmq_read('my_queue', 30, 1, '{"type": "A"}');`

Thank you for considering this feature request.

Neptune650:
/claim #288
Closes #288

This adds a new parameter for read, adds tests and docs for it and creates a migration file due to the changed function signature.

However `pgmq.archive()` and `pgmq.delete()` are unchanged because they already receive a msg_id parameter, so there's no need to filter with a conditional, and would cause issues if the conditional isn't true.

### DIFF
--- a/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
+++ b/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
@@ -1,0 +1,98 @@
+-- read
+-- reads a number of messages from a queue, setting a visibility timeout on them
+DROP FUNCTION pgmq.read(TEXT, INTEGER, INTEGER);
+CREATE OR REPLACE FUNCTION pgmq.read(
+    queue_name TEXT,
+    vt INTEGER,
+    qty INTEGER,
+    conditional JSONB DEFAULT '{}'
+)
+RETURNS SETOF pgmq.message_record AS $$
+DECLARE
+    sql TEXT;
+    qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+BEGIN
+    sql := FORMAT(
+        $QUERY$
+        WITH cte AS
+        (
+            SELECT msg_id
+            FROM pgmq.%I
+            WHERE vt <= clock_timestamp() AND message @> %I
+            ORDER BY msg_id ASC
+            LIMIT $1
+            FOR UPDATE SKIP LOCKED
+        )
+        UPDATE pgmq.%I m
+        SET
+            vt = clock_timestamp() + %L,
+            read_ct = read_ct + 1
+        FROM cte
+        WHERE m.msg_id = cte.msg_id
+        RETURNING m.msg_id, m.read_ct, m.enqueued_at, m.vt, m.message;
+        $QUERY$,
+        qtable, conditional, qtable, make_interval(secs => vt)
+    );
+    RETURN QUERY EXECUTE sql USING qty;
+END;
+$$ LANGUAGE plpgsql;
+
+---- read_with_poll
+---- reads a number of messages from a queue, setting a visibility timeout on them
+DROP FUNCTION pgmq.read_with_poll(TEXT, INTEGER, INTEGER, INTEGER, INTEGER);
+CREATE OR REPLACE FUNCTION pgmq.read_with_poll(
+    queue_name TEXT,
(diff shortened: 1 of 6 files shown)


Neptune650:
@ChuckHend Fixed the tests as well here, please review.

pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql
```diff
@@ -0,0 +1,98 @@
...
+    sql TEXT;
+    qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+BEGIN
+    sql := FORMAT(
+        $QUERY$
+        WITH cte AS
+        (
+            SELECT msg_id
+            FROM pgmq.%I
+            WHERE vt <= clock_timestamp() AND message @> %I
```
ChuckHend, line 21: Looks like this is in the right direction. Although I'm not sure `%I` will work for jsonb type, it might need to be a `%L`?

https://github.com/tembo-io/pgmq/actions/runs/11430681669/job/31799398848?pr=322

Also, if message is `null` then this will make `pgmq.read()` unable to return any messages since the DEFAULT value of the `conditional` parameter is `'{}'`

ChuckHend: pgmq.read() no longer fails when I change that `%I` to a `%L`, but it is unable to retrieve when message value is `null`.

```
postgres=# select pgmq.create('test');
 create 
--------
 
(1 row)

postgres=# select pgmq.send('test', null);
 send 
------
    1
(1 row)

postgres=# select pgmq.read('test', 1, 1);
 read 
------
(0 rows)

postgres=# select * from pgmq.q_test;
 msg_id | read_ct |          enqueued_at          |             vt              | message 
--------+---------+-------------------------------+-----------------------------+---------
      1 |       0 | 2024-10-20 20:56:14.760672-04 | 2024-10-20 20:56:14.7619-04 | 
(1 row)
```

Neptune650: @ChuckHend Should be fixed now.

Neptune650:
@ChuckHend Any idea why the tests are failing now? You mentioned using %L worked for you.

pgmq-extension/test/sql/base.sql
```diff
@@ -29,6 +29,9 @@ SELECT * from pgmq.send('test_default_queue', '{"hello": "world"}');
 \set msg_id 1
 SELECT msg_id = :msg_id FROM pgmq.read('test_default_queue', 2, 1);
 
+-- read message using conditional
+SELECT msg_id = :msg_id FROM pgmq.read('test_default_queue', 2, 1, '{"hello": "world"}');
+
```
ChuckHend, line 34: @Neptune650 , maybe this test fails because the statement above sets VT to 2 seconds, which means the message is probably invisible when we try to read on this line?

Neptune650: You were correct about this, thanks.

Neptune650:
@ChuckHend All the tests have been fixed, could you please review, and if appropiate, merge this PR and #320.

v0idpwn (approved):
Looks great to me.

v0idpwn (commented):
I think it would be good to write a guide on how to use this, and how to index it, @ChuckHend, but it's probably out of scope for this PR.

Neptune650:
> I think it would be good to write a guide on how to use this, and how to index it, @ChuckHend, but it's probably out of scope for this PR.

I did add documentation explaining the usage of the parameter here which I figured would be enough for this PR.

ChuckHend (approved):

ChuckHend:
Thank you @Neptune650!

Neptune650:
> Thank you @Neptune650!

Thanks to you for reviewing and merging!

snosratiershad:
@ChuckHend, @Neptune650, @v0idpwn I think this feature needs to be integrated in clients too.

PR #333, opened 2024-10-28
Add Python extension implementation for send with timestamp

Neptune650:
Needs #320 first to be merged.

This adds the Python extension part of send with timestamp, and adds tests for it.

### DIFF
--- a/tembo-pgmq-python/tembo_pgmq_python/async_queue.py
+++ b/tembo-pgmq-python/tembo_pgmq_python/async_queue.py
@@ -155,45 +155,79 @@ class PGMQueue:
         return queues
 
     @transaction
-    async def send(self, queue: str, message: dict, delay: int = 0, conn=None) -> int:
+    async def send(self, queue: str, message: dict, delay: int = 0, tz: str = None, conn=None) -> int:
         """Send a message to a queue."""
-        self.logger.debug(f"send called with queue='{queue}', message={message}, delay={delay}, conn={conn}")
+        self.logger.debug(f"send called with queue='{queue}', message={message}, delay={delay}, tz={tz}, conn={conn}")
         if conn is None:
             async with self.pool.acquire() as conn:
-                return await self._send_internal(queue, message, delay, conn)
+                return await self._send_internal(queue, message, delay, tz, conn)
         else:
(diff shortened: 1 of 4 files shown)


Neptune650:
Also - naturally - the tests will fail until #320 is merged.

tavallaie:
> Also - naturally - the tests will fail until #320 is merged.

Thanks for your contribution! 
let's get #320 merge to continue from there.

tembo-pgmq-python/tembo_pgmq_python/async_queue.py
```diff
@@ -155,45 +155,79 @@ async def _list_queues_internal(self, conn):
         return queues
 
     @transaction
-    async def send(self, queue: str, message: dict, delay: int = 0, conn=None) -> int:
+    async def send(self, queue: str, message: dict, delay: int = 0, tz: str = None, conn=None) -> int:
         """Send a message to a queue."""
-        self.logger.debug(f"send called with queue='{queue}', message={message}, delay={delay}, conn={conn}")
+        self.logger.debug(f"send called with queue='{queue}', message={message}, delay={delay}, tz={tz}, conn={conn}")
```
ChuckHend, line 160: The python logging module is optimized for `%s`. Basically we don't need format and interpolate the log string unless we're actually going to log this message. So IMO, we should use `%s` on all calls to the python logging.

https://docs.python.org/3/howto/logging.html#optimization

ChuckHend: Also, I realize the codebase is already using fstring all over the place, so I'm not asking us to change this in this PR.

Neptune650: @ChuckHend No issue then?

ChuckHend: Correct,  not an issue

ChuckHend:
@tavallaie , @Neptune650 , I pushed a few commits. I think this is ready now.

ChuckHend (approved):

tavallaie:
I think everything was good, when will you publishing it on pypi?

ChuckHend:
It was already published to pypi 😀

Issue #238, opened 2024-05-29
Installation guide for multiple platforms


v0idpwn:
```[tasklist]
### Tasks
- [ ] Guide for PGXN
- [ ] Guide for Windows
- [ ] Guide for Mac
```

ChuckHend:
@v0idpwn , I think now that this is pl/pgsql we would just need to add instructions for windows? `make install` pretty much does everything now.

nyarly:
For reference, I bounced off of pgmq because all the documentation for getting started was "run this Docker image" - I already have a postgresql instance, and I'm not interested in dockerizing it.

Unless there's clear, standalone install instructions for pgmq, I'm not interested.

theory:
Probably worth adding a few options to the README:

``` sh
pgxn install pgmq
```

Or:

``` sh
curl -LO https://api.pgxn.org/dist/pgmq/1.4.2/pgmq-1.4.2.zip
unzip pgmq-1.4.2.zip
cd pgmq-1.4.2
make
make install
```

nyarly:
Okay, but what's `pgxn`?

theory:
> PGXN, the PostgreSQL Extension network, is a central distribution system for open-source PostgreSQL extension libraries.

— [About PGXN](https://pgxn.org/about/)

PR #323, opened 2024-10-19
Add Installation Guide  for PGMQ

Saturn225:
Closes #238
/claim #238

PR #335, opened 2024-10-31
Add build instructions for PGMQ

Neptune650:
Closes https://github.com/tembo-io/pgmq/issues/238
/claim https://github.com/tembo-io/pgmq/issues/238

@ChuckHend @theory I created this PR to solve the issues pointed out in #323, since the original creator of the pull request hasn't been active.

I changed the MinGW requirement to MSYS2, since it's the same but the latter is pre-packaged and much easier to work with for new users.

And removed the macOS steps because @theory has pointed out we don't really need these.


Saturn225:
@Neptune650 Thanks for taking effort for following the issue. I have been jut got busy in other works on recent week. Hopefully I am free now and push my PR to end :)

Neptune650:
@Saturn225 Sounds good to me

PR #351, opened 2024-12-06
Add Supabase to "Who uses pgmq?"

jumski:
Supabase just released Supabase Queues which uses pgmq under the hood (yay!). Is it ok to count them in?

:partying_face:


ChuckHend:
@olirice , any issue if we put this on the readme? I think it looks nice 😀

olirice:
yes, absolutely :pray:
thanks for checking

v0idpwn (approved):

PR #352, opened 2024-12-06
Add created sequence object to the extension

za-arthur:
Without adding the sequence pg_dump crashes. This happens because the sequence's table is ignored during dumping objects, unlike the sequence. It leads to inconsistency in pg_dump's metadata of objects.

Here is the backtrace from Postgres 15.10:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x000000010001d140 pg_dump`dumpTable [inlined] dumpSequence(fout=0x000000015b6049e0, tbinfo=0x00000001600c0040) at pg_dump.c:16978:15 [opt]
   16975                                                          fmtQualifiedDumpable(owning_tab));
   16976                        appendPQExpBuffer(query,
   16977                                                          "ALTER COLUMN %s ADD GENERATED ",
-> 16978                                                          fmtId(owning_tab->attnames[tbinfo->owning_col - 1]));
   16979                        if (owning_tab->attidentity[tbinfo->owning_col - 1] == ATTRIBUTE_IDENTITY_ALWAYS)
   16980                                appendPQExpBufferStr(query, "ALWAYS");
   16981                        else if (owning_tab->attidentity[tbinfo->owning_col - 1] == ATTRIBUTE_IDENTITY_BY_DEFAULT)
warning: pg_dump was compiled with optimization - stepping may behave oddly; variables may not be available.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x000000010001d140 pg_dump`dumpTable [inlined] dumpSequence(fout=0x000000015b6049e0, tbinfo=0x00000001600c0040) at pg_dump.c:16978:15 [opt]
    frame #1: 0x000000010001d04c pg_dump`dumpTable(fout=0x000000015b6049e0, tbinfo=0x00000001600c0040) at pg_dump.c:15102:4 [opt]
    frame #2: 0x00000001000061a0 pg_dump`main [inlined] dumpDumpableObject(fout=0x000000015b6049e0, dobj=0x00000001600c0040) at pg_dump.c:10045:4 [opt]
    frame #3: 0x00000001000060e4 pg_dump`main(argc=<unavailable>, argv=<unavailable>) at pg_dump.c:912:3 [opt]
    frame #4: 0x00000001993ac274 dyld`start + 2840
```

### DIFF
--- a/pgmq-extension/sql/pgmq.sql
+++ b/pgmq-extension/sql/pgmq.sql
@@ -588,6 +588,7 @@ CREATE FUNCTION pgmq.drop_queue(queue_name TEXT)
 RETURNS BOOLEAN AS $$
 DECLARE
     qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+    qtable_seq TEXT := qtable || '_msg_id_seq';
     fq_qtable TEXT := 'pgmq.' || qtable;
     atable TEXT := pgmq.format_table_name(queue_name, 'a');
     fq_atable TEXT := 'pgmq.' || atable;
@@ -605,6 +606,13 @@ BEGIN
         ALTER EXTENSION pgmq DROP TABLE pgmq.%I
         $QUERY$,
         qtable
+    );
+
+    EXECUTE FORMAT(
+        $QUERY$
+        ALTER EXTENSION pgmq DROP SEQUENCE pgmq.%I
+        $QUERY$,
+        qtable_seq
     );
 
     EXECUTE FORMAT(
@@ -687,6 +695,7 @@ CREATE FUNCTION pgmq.create_non_partitioned(queue_name TEXT)
 RETURNS void AS $$
 DECLARE
   qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+  qtable_seq TEXT := qtable || '_msg_id_seq';
   atable TEXT := pgmq.format_table_name(queue_name, 'a');
 BEGIN
   PERFORM pgmq.validate_queue_name(queue_name);
@@ -722,6 +731,7 @@ BEGIN
 
   IF NOT pgmq._belongs_to_pgmq(qtable) THEN
       EXECUTE FORMAT('ALTER EXTENSION pgmq ADD TABLE pgmq.%I', qtable);
+      EXECUTE FORMAT('ALTER EXTENSION pgmq ADD SEQUENCE pgmq.%I', qtable_seq);
   END IF;
 
   IF NOT pgmq._belongs_to_pgmq(atable) THEN
@@ -758,6 +768,7 @@ CREATE FUNCTION pgmq.create_unlogged(queue_name TEXT)
 RETURNS void AS $$
 DECLARE
   qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+  qtable_seq TEXT := qtable || '_msg_id_seq';
   atable TEXT := pgmq.format_table_name(queue_name, 'a');
 BEGIN
   PERFORM pgmq.validate_queue_name(queue_name);
@@ -792,6 +803,7 @@ BEGIN
 
   IF NOT pgmq._belongs_to_pgmq(qtable) THEN
       EXECUTE FORMAT('ALTER EXTENSION pgmq ADD TABLE pgmq.%I', qtable);
+      EXECUTE FORMAT('ALTER EXTENSION pgmq ADD SEQUENCE pgmq.%I', qtable_seq);
   END IF;
 
   IF NOT pgmq._belongs_to_pgmq(atable) THEN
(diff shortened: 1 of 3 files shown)


v0idpwn (commented):
This sounds right to me, but it raises two things:
1. Should it be added somewhere else?
2. We must add this to a version upgrade script too, so that when updating pgmq, people with an older version get their data corrected

za-arthur:
Also I forgot to DROP the sequence within `pgmq.drop_queue`.

BTW I wonder why queue tables are added to the extension? Would it make sense to allow users to dump those tables?

pgmq-extension/sql/pgmq.sql
```diff
@@ -653,6 +653,7 @@ BEGIN
 
   IF NOT pgmq._belongs_to_pgmq(qtable) THEN
       EXECUTE FORMAT('ALTER EXTENSION pgmq ADD TABLE pgmq.%I', qtable);
+      EXECUTE FORMAT('ALTER EXTENSION pgmq ADD SEQUENCE pgmq.%I_msg_id_seq', qtable);
```
olirice, line 656: `%I_msg_id_seq` may be escaped as `"some-table"_msg_id_seq` in some cases

ChuckHend:
> BTW I wonder why queue tables are added to the extension? Would it make sense to allow users to dump those tables?

The tables are added to the extension so that when you remove the extension from your database, the objects created by the extension (including the queue tables) will also be removed. The [detach_archive](https://github.com/tembo-io/pgmq/blob/57d8a9b4dbe9cac82734f5387f04aa7ac0050d0b/pgmq-extension/sql/pgmq.sql#L434-L435) function was added as a way to keep the archive table.

pgmq-extension/sql/pgmq.sql
za-arthur: I fixed this by using a separate variable `qtable_seq` which is passed to `FORMAT`.

za-arthur:
I pushed latest change which
- adds `ADD SEQUENCE`/`DROP SEQUENCE` to all relevant functions `pgmq.drop_queue`, `pgmq.create_non_partitioned`, `pgmq.create_unlogged`, `pgmq.create_partitioned`
- adds the migration script `pgmq--1.5.0--1.5.1.sql` which redefines the functions and adds relevant sequences to the extension

pgmq-extension/pgmq.control
```diff
@@ -1,5 +1,5 @@
 comment = 'A lightweight message queue. Like AWS SQS and RSMQ but on Postgres.'
-default_version = '1.5.0'
+default_version = '1.5.1'
```
ChuckHend, line 2: Let's plan for this to go out with v1.5.0 since it has not yet been released.

pgmq-extension/sql/pgmq--1.5.0--1.5.1.sql
```diff
@@ -0,0 +1,415 @@
...
+    -- Get all queue tables which belongs to the extension
+    FOR qtable IN
+        SELECT c.relname
+        FROM pg_catalog.pg_depend d
+        JOIN pg_catalog.pg_extension e ON e.oid = d.refobjid AND e.extname = 'pgmq'
+        JOIN pg_catalog.pg_namespace n ON n.oid = e.extnamespace AND n.nspname = 'pgmq'
+        JOIN pg_catalog.pg_class c ON c.oid = d.objid AND c.relname LIKE 'q_%' AND c.relkind IN ('r', 'p')
+        WHERE d.refclassid = 'pg_extension'::regclass AND d.deptype = 'e' AND d.classid = 'pg_class'::regclass
+    LOOP
+        qtable_seq := qtable || '_msg_id_seq';
```
ChuckHend, line 17: the `pgmq.meta` table will contain all the queue names that pgmq is aware of.

za-arthur:
@ChuckHend I returned the version back to `1.5.0` and the migration script uses `pgmq.meta` now. I used more complex query before to make sure that the migration is idempotent. But I guess this might be an issue only if a user added sequences manually.

BTW it seems that the migration script skips `1.4.5`. Is this intentional? It looks like a typo and it might fail during upgrading from `1.4.5` to `1.5.0`.

ChuckHend:
@za-arthur, correct we will need an upgrade script for 1.4.5 to 1.5.0 before we release 1.5.0, thank you for bringing that up. I think we may need to rename https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq--1.4.4--1.5.0.sql to `..1.4.5--1.5.0.sql` then put the changes from this PR into that migration.

za-arthur:
@ChuckHend I created a PR to rename the script:
https://github.com/tembo-io/pgmq/pull/355

za-arthur:
Rebased the PR on latest changes.

ChuckHend (approved):
Thank you!

PR #356, opened 2024-12-11
drop pg12 support and publish pg17 to pgt.dev

ChuckHend:


theory (approved):
Not sure why 12 needs to be dropped, but not a big deal. Curious how this triggers a trunk publish.

ChuckHend:
pg12 is end of life right?
https://www.postgresql.org/support/versioning/

Maybe the PR title is misleading. This wont trigger a trunk publish. But the next next we do a release it will, and this adds pg17 into that build matrix.

PR #358, opened 2024-12-13
docs to clarify read_with_poll param qty

### LINKED ISSUE #357: python client read_with_poll not working as expected
```
read_messages = queue.read_with_poll(
    "my_queue", vt=30, qty=500, max_poll_seconds=1000, poll_interval_ms=100
)
for message in read_messages:
    print(message)
```
For this scenario what we want should read_messages should be length=500 or the `read_with_poll` should be blocked util 1000 seconds.

But what I got was, it immediately print all 20 messages(all message to my_queue) and then exit.

elvizlai:
fix #357

### DIFF
--- a/tembo-pgmq-python/README.md
+++ b/tembo-pgmq-python/README.md
@@ -135,7 +135,7 @@ for message in read_messages:
     print(message)
 ```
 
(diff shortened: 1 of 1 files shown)


ChuckHend (approved):
Thank you!

PR #360, opened 2024-12-13
fix: python logging

### LINKED ISSUE #359: python: Using logging.basicConfig Causes Loggers in Other Components to Behave Unexpectedly
When logging.basicConfig is called in a Python application, it can unintentionally override the logging configurations of other components, leading to unexpected behavior.

https://github.com/tembo-io/pgmq/blob/6826e182ae1ca913fefe10aba44c5fd0ea070bec/tembo-pgmq-python/tembo_pgmq_python/queue.py#L42-L52

https://github.com/tembo-io/pgmq/blob/6826e182ae1ca913fefe10aba44c5fd0ea070bec/tembo-pgmq-python/tembo_pgmq_python/async_queue.py#L47-L57

Maybe we should fix with:
using logging.getLogger(__name__) with specify level and handler

elvizlai:
fix #359

### DIFF
--- a/tembo-pgmq-python/pyproject.toml
+++ b/tembo-pgmq-python/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "tembo-pgmq-python"
-version = "0.9.0"
+version = "0.9.1"
 description = "Python client for the PGMQ Postgres extension."
 authors = ["Adam Hendel <adam@tembo.io>"]
 license = "PostgreSQL"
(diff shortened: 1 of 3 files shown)


ChuckHend:
@tavallaie could you take a look?

tavallaie (approved):

ChuckHend (approved):

tavallaie:
It is a really good idea, I should do it in first place.

PR #363, opened 2024-12-18
fix 1.5.0 migration per issue #362

### LINKED ISSUE #362: Migration for 1.5.0 broken
In https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq--1.4.5--1.5.0.sql

- The actual implementation of `send` (https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq--1.4.5--1.5.0.sql#L610) should come before the simpler overloads at https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq--1.4.5--1.5.0.sql#L565
- The actual implementation of `send_batch` (https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq--1.4.5--1.5.0.sql#L678) should come before the simpler overloads at https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq--1.4.5--1.5.0.sql#L633

WIthout this change applying it fails with

```
SQL State  : 42883
Error Code : 0
Message    : ERROR: function pgmq.send(text, jsonb, unknown, timestamp with time zone) does not exist
  Hint: No function matches the given name and argument types. You might need to add explicit type casts.
```

and

```
SQL State  : 42883
Error Code : 0
Message    : ERROR: function pgmq.send_batch(text, jsonb[], unknown, timestamp with time zone) does not exist
  Hint: No function matches the given name and argument types. You might need to add explicit type casts.
```

respectively.

What really should happen here, is that all the migrations should be automatically integration tested  as part of the release process to prevent this from occurring again.

ChuckHend:
closes #362

### DIFF
--- a/pgmq-extension/pgmq.control
+++ b/pgmq-extension/pgmq.control
@@ -1,5 +1,5 @@
 comment = 'A lightweight message queue. Like AWS SQS and RSMQ but on Postgres.'
-default_version = '1.5.0'
+default_version = '1.5.1'
 module_pathname = '$libdir/pgmq'
 schema = 'pgmq'
 relocatable = false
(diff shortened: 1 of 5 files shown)


ChuckHend:
closing until we get more information in #362

Issue #378, opened 2025-02-05
GCP Support


shinzui:
I'm not sure if opening an issue for this is appropriate, but for those interested in using the pgmq extension on GCP Cloud SQL, I wanted to share this.

`pgmq` would be usable on GCP Cloud SQL if it supported something like AWS's `pg_tle`. I've submitted a [request](https://issuetracker.google.com/issues/393315125) to GCP to add support for pg_tle or an equivalent feature, but for them to prioritize it, they need to see enough interest from users.

If you're on GCP and would benefit from this, please consider [upvoting](https://issuetracker.google.com/issues/393315125) the request to help increase visibility

jumski:
Isn't pgmq really pure SQL? There is some rust code for extension, but I think it is pgrx leftover.

You can just grab the migration and apply it to your schema.

ChuckHend:
We just merged https://github.com/tembo-io/pgmq/pull/379 which will make it much easier to create PGMQ's SQL objects (not as an extension). We don't have any tests or docs written for it yet, but yes you should be able to execute the raw SQL into any Postgres instance.

shinzui:
Great! #379 takes care of this.

Issue #373, opened 2025-01-10
Add pgmq extension to Microsoft Azure Postgres flexible server (upvote link)


ps30:
Please click this link and **upvote** & follow to help Microsoft to add/approve pgmq as a supported Postgres extension:

[Add extension pgmq - https://feedback.azure.com/d365community/idea/9d5c7e3b-f35f-ef11-a4e6-6045bdbe5235?q=pgmq](https://feedback.azure.com/d365community/idea/9d5c7e3b-f35f-ef11-a4e6-6045bdbe5235?q=pgmq)

Raw text link: https://feedback.azure.com/d365community/idea/9d5c7e3b-f35f-ef11-a4e6-6045bdbe5235?q=pgmq

wahmedswl:
When this is SQL only; clients have logic built-in to invoke how it should be. So, if we just create schema in postgres; it should work, isn't it?

Thanks

ChuckHend:
> So, if we just create schema in postgres; it should work, isn't it?

I've heard of some users using the SQL directly to install the objects in this extension. However, the migration scripts are generally not executable directly and are intended to be run with the `alter extension pgmq update` command. I have not done this myself.

wahmedswl:
@ChuckHend When it's all sql based; no dependency etc then it would be great for compatibility to make it sql only. It would be great if schema is maintained as a single sql file for these cases. 

Thanks

ChuckHend:
> [@ChuckHend](https://github.com/ChuckHend) When it's all sql based; no dependency etc then it would be great for wide compatibility to make it sql only. It would be great if schema is maintained as a single sql file for these cases.
> 
> Thanks

If you want to help out, https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq.sql contains all the functions and DDL (except the create schema statement, I think). For anyone looking to run this as SQL-only, that is the best place to start. Help is very much welcome here!

Issue #347, opened 2024-11-18
Don't assume to be installed as an extension


axelfontaine:
Now that the migration to pure PL/pgSQL is complete, it is very easy to run `pgmq` on any PostgreSQL instance by simply:

1. Creating a `pgmq` schema
2. Executing `pgmq.sql`

... or so it should be!

Unfortunately it doesn't quite work out of the box and the following changes are necessary:

```patch
Index: C:/Users/axel/AppData/Local/Temp/pgmq.sql
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/C:/Users/axel/AppData/Local/Temp/pgmq.sql b/sprinters/src/main/resources/flyway/pgmq/pgmq--1.4.5.sql
rename from C:/Users/axel/AppData/Local/Temp/pgmq.sql
rename to sprinters/src/main/resources/flyway/pgmq/pgmq--1.4.5.sql
--- a/C:/Users/axel/AppData/Local/Temp/pgmq.sql	
+++ b/sprinters/src/main/resources/flyway/pgmq/pgmq--1.4.5.sql	(date 1731965677611)
@@ -390,16 +390,6 @@
 END
 $$ LANGUAGE plpgsql;
 
--- unassign archive, so it can be kept when a queue is deleted
-CREATE FUNCTION pgmq."detach_archive"(queue_name TEXT)
-RETURNS VOID AS $$
-DECLARE
-  atable TEXT := pgmq.format_table_name(queue_name, 'a');
-BEGIN
-  EXECUTE format('ALTER EXTENSION pgmq DROP TABLE pgmq.%I', atable);
-END
-$$ LANGUAGE plpgsql;
-
 -- pop a single message
 CREATE FUNCTION pgmq.pop(queue_name TEXT)
 RETURNS SETOF pgmq.message_record AS $$
@@ -468,20 +458,6 @@
     atable TEXT := pgmq.format_table_name(queue_name, 'a');
     fq_atable TEXT := 'pgmq.' || atable;
 BEGIN
-    EXECUTE FORMAT(
-        $QUERY$
-        ALTER EXTENSION pgmq DROP TABLE pgmq.%I
-        $QUERY$,
-        qtable
-    );
-
-    EXECUTE FORMAT(
-        $QUERY$
-        ALTER EXTENSION pgmq DROP TABLE pgmq.%I
-        $QUERY$,
-        atable
-    );
-
[...]

Issue #205, opened 2024-05-09
Question: Does this work on hosted providers?


dwelch2344:
I'm just discovering the awesomeness that is Tembo via this project, so forgive me if this is a stupid question. 

Is this something that can run on top of any "standard" PG instance out there? ie AWS RDS, GCP CloudSQL, Heroku, etc.

ChuckHend:
If the cloud providers add it to their supported extensions, then there is no reason it cannot. I think it requires filing a support ticket to get a new extension added. The work discussed in https://github.com/tembo-io/pgmq/issues/199 would make it a bit more portable (re-writing certain parts from Rust into pl/pgsql).

I would say it is generally very difficult to get new extension added to most Postgres providers out there.

ConProgramming:
Could this run as Pl/rust for a trusted language extension? https://aws.amazon.com/blogs/database/build-high-performance-functions-in-rust-on-amazon-rds-for-postgresql/

ChuckHend:
@ConProgramming, I have not built with PL/Rust so I cannot say for sure. Partitioned queue's depend on pg_partman, which I am almost certain will not run as PL/Rust, however most cloud providers already provide that extension.

dwelch2344:
Oooo ive used a number of other pl/X langs before, but haven't yet dabbled in rust. Happy to take a stab at this. If anyone wants to collaborate, lmk

ChuckHend:
> Oooo ive used a number of other pl/X langs before, but haven't yet dabbled in rust. Happy to take a stab at this. If anyone wants to collaborate, lmk

All the extension functions are [now in pl/pgsql ](https://github.com/tembo-io/pgmq/blob/main/src/sql_src.sql). The build/packaging system is still in Rust though. We will most likely be moving that over to pgxs soon.

ConProgramming:
> > Oooo ive used a number of other pl/X langs before, but haven't yet dabbled in rust. Happy to take a stab at this. If anyone wants to collaborate, lmk
> 
> All the extension functions are [now in pl/pgsql ](https://github.com/tembo-io/pgmq/blob/main/src/sql_src.sql). The build/packaging system is still in Rust though. We will most likely be moving that over to pgxs soon.

Ah wasn't aware extensions functions are now in pl/pgsql, that means this should work on hosted providers without a problem yes? Just running the [pl/pgsql](https://github.com/tembo-io/pgmq/blob/main/src/sql_src.sql) on a server?

[...]

PR #379, opened 2025-02-09
Enable PGMQ to run without installing an extension

axelfontaine:
This pull request enables the installation of PGMQ using a simple `psql` command:

```
psql < pgmq.sql
```

It does NOT affect the existing functionality, nor does it impact PGMQ's ability to be installed as an extension.

This addresses the following issues directly and inderectly:
- #205
- #347
- #373 
- #378

To achieve this the following was changed:
- The `pgmq` schema gets created if it doesn't already exist
- The actual implementation of `send` and `send_batch` are moved higher up the SQL file so they can be found when referenced by the convenience functions with less arguments
- A new `_extension_exists` function is introduced to check the existance of arbitrary extensions
- The implementation of `_ensure_pg_partman_installed` has been retrofitted to use the new `_extension_exists` function.
- All references to the `pgmq` extension have been guarded with `_extension_exists('pgmq')`

### DIFF
--- a/pgmq-extension/sql/pgmq.sql
+++ b/pgmq-extension/sql/pgmq.sql
@@ -1,8 +1,10 @@
 ------------------------------------------------------------
 -- Schema, tables, records, privileges, indexes, etc
 ------------------------------------------------------------
--- We don't need to create the `pgmq` schema because it is automatically
--- created by postgres due to being declared in extension control file
+-- When installed as an extension, we don't need to create the `pgmq` schema
+-- because it is automatically created by postgres due to being declared in
+-- the extension control file
+CREATE SCHEMA IF NOT EXISTS pgmq;
 
 -- Table where queues and metadata about them is stored
 CREATE TABLE pgmq.meta (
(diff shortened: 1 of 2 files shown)


ChuckHend:
This is awesome, thank you for doing this @axelfontaine. I will look it over in the next few days.

ChuckHend:
image builds are going to fail for PR coming from an external repo (we still need a better way to configure that in CI), so we can ignore those failures.

v0idpwn:
This is great. I think we should have tests run on it, though.

axelfontaine:
@ChuckHend Versioning is not something PGMQ should be concerned with. I use Flyway, but other similar tools will also do the job.

Upgrading means copying the PGMQ upgrade script into one's project and giving it the next version according to Flyway's (or whatever other tool's) naming convention.

Or simply applying it manually using psql. 

**PGMQ should exclusively deal with queueing. Let other specialized tools handle the versioning.**

ChuckHend:
Understood. I was thinking of this from the perspective of providing a similar experience to what pgmq has as a Postgres extension, but as stand-alone functionality. It is helpful for me, when troubleshooting, to know which version of the pgmq project folks are running, and versioning is already part of our release process. We don't need to block this PR on versioning though.

I do think we need to have some sort of test in place for executing the migration sql files for stand-alone pgmq though, so that we don't break it for all the non extension users. @axelfontaine what do you think?

axelfontaine:
> It is helpful for me, when troubleshooting, to know which version of the pgmq project folks are running, and versioning is already part of our release process.

This can be tackled by providing a `pgmq.version()` function that returns a hardcoded string with the version. Simple and effective, without getting tangled up in the complexity web of a full migration solution. I worked on Flyway for 10 years before selling in in 2019. Trust me, I know what I am talking about 😉

> I do think we need to have some sort of test in place for executing the migration sql files for stand-alone pgmq though, so that we don't break it for all the non extension users. @axelfontaine what do you think?

Agreed. I actually think all tests should be decoupled from GHA and also easily runnable locally at least on Linux, but ideally on all platforms (Win/Mac/Linux). I see that as being separate from this PR as well though.

v0idpwn (approved):

PR #383, opened 2025-02-21
Change pop() and set_vt() to use clock_timestamp() instead of now()

### LINKED ISSUE #367: Consistently use clock_timestamp() instead of now() when working with vt
Most of the time, pgmq uses `clock_timestamp()` when setting or comparing `vt`.  This is the case for the following functions:
* read
* read_with_poll
* send
* send_batch

But these functions use `now()`:
* [pop](https://github.com/tembo-io/pgmq/blob/05f27a711feb12f131b59ce3ff3bd732adfa6240/pgmq-extension/sql/pgmq.sql#L525)
* [set_vt](https://github.com/tembo-io/pgmq/blob/05f27a711feb12f131b59ce3ff3bd732adfa6240/pgmq-extension/sql/pgmq.sql#L551)

Is there a reason for the difference?

Depending on which is used, the behavior can be different when used inside a transaction, which might be unexpected.  I ran into this while testing something and it surprised me until I looked at the code.

Here is an example showing what I'm talking about:  
Inside a transaction, send a message to the queue.  You can't pop the message, but you can read it.
```
postgres=# begin transaction;
BEGIN
postgres=*# select pgmq.send('my_queue', '{"foo": "bar1"}');
 send 
------
   38
(1 row)

postgres=*# select pgmq.pop('my_queue');
 pop 
-----
(0 rows)

postgres=*# select pgmq.read('my_queue', 30, 1);
                                             read                                             
----------------------------------------------------------------------------------------------
 (37,2,"2024-12-20 19:44:53.600822+00","2024-12-20 19:46:10.624711+00","{""foo"": ""bar1""}")
(1 row)

postgres=*# commit;
COMMIT
```

If there isn't a reason for them to be different, I think pop and set_vt should use `clock_timestamp()`, to be consistent with the other functions.

brianpursley:
Fixes #367

### DIFF
--- a/pgmq-extension/pgmq.control
+++ b/pgmq-extension/pgmq.control
@@ -1,5 +1,5 @@
 comment = 'A lightweight message queue. Like AWS SQS and RSMQ but on Postgres.'
-default_version = '1.5.0'
+default_version = '1.5.1'
 module_pathname = '$libdir/pgmq'
 schema = 'pgmq'
 relocatable = false
--- a/pgmq-extension/sql/pgmq--1.5.0--1.5.1.sql
+++ b/pgmq-extension/sql/pgmq--1.5.0--1.5.1.sql
@@ -1,0 +1,49 @@
+-- pop a single message
+CREATE OR REPLACE FUNCTION pgmq.pop(queue_name TEXT)
+RETURNS SETOF pgmq.message_record AS $$
+DECLARE
+    sql TEXT;
+    result pgmq.message_record;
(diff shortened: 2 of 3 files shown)


brianpursley:
Let me know if you want tests for this

jumski:
Looks good @brianpursley, thanks for fixing it!

v0idpwn (approved):

ChuckHend (approved):
Thank you!

PR #385, opened 2025-03-26
pgmq.drop_queue return false when queue does not exist

### LINKED ISSUE #275: `pgmq.drop_queue` issues when queue doesn't exist
Hello,

We have this code (https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/sql/pgmq.sql#L429) for dropping a queue:
```sql
ALTER EXTENSION pgmq DROP TABLE pgmq.q_%s;

ALTER EXTENSION pgmq DROP TABLE pgmq.a_%s;

DROP TABLE IF EXISTS pgmq.q_%s;

DROP TABLE IF EXISTS pgmq.a_%s;
```
Currently, when I drop a non-existing queue, the first statement fails with an error:
```
ERROR:  relation "pgmq.q_test" does not exist
CONTEXT:  SQL statement "
        ALTER EXTENSION pgmq DROP TABLE pgmq.q_test
        "
```

From PostgreSQL [docs](https://www.postgresql.org/docs/current/sql-alterextension.html) on `DROP member_object`:

> This form removes a member object from the extension. This is mainly useful in extension update scripts. The object is not dropped, only disassociated from the extension.

We're in a user-called function here, not in an update script. Is there a point of altering the extension itself with queue/archive tables? Please look at 2 other SQL statements: we are careful here not to throw an error using the `IF EXISTS`. There is no `IF EXISTS` counterpart for `ALTER EXTENSION ... DROP TABLE` so maybe we just get rid of that `ALTER EXTENSION` altogether?

In fact, this causes even more problems. For example:
```sql
postgres=# select pgmq.create('test');
 create
--------

(1 row)

postgres=# select pgmq.detach_archive('test');
 detach_archive
----------------

(1 row)

postgres=# select pgmq.drop_queue('test');
ERROR:  table pgmq.a_test is not a member of extension "pgmq"
CONTEXT:  SQL statement "
        ALTER EXTENSION pgmq DROP TABLE pgmq.a_test
        "
PL/pgSQL function pgmq.drop_queue(text,boolean) line 10 at EXECUTE
```

`pgmq` is supposed to be used in highly concurrent environments - I think the less exceptions like that are thrown, the easier it is to maintain the software that uses it.

ChuckHend:
Updates `pgmq.drop_queue()` does not fail when attempting to drop a queue that does not exist. Instead it will just return `false` and creates a notice. Closes #275.

The implementation simply checks if the queue table exists prior to attempting to delete it.

Previous behavior:

```text
postgres=# select pgmq.drop_queue('dne');
ERROR:  relation "pgmq.q_dne" does not exist
CONTEXT:  SQL statement "
            ALTER EXTENSION pgmq DROP TABLE pgmq.q_dne
            "
PL/pgSQL function pgmq.drop_queue(text) line 18 at EXECUTE
```

New behavior

```text
postgres=# select pgmq.drop_queue('dne');
NOTICE:  pgmq queue `dne` does not exist
 drop_queue 
------------
 f
(1 row)
```

### DIFF
--- a/pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
+++ b/pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
@@ -1,0 +1,90 @@
+DROP FUNCTION pgmq.drop_queue(TEXT);
+CREATE FUNCTION pgmq.drop_queue(queue_name TEXT)
+RETURNS BOOLEAN AS $$
+DECLARE
+    qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+    qtable_seq TEXT := qtable || '_msg_id_seq';
+    fq_qtable TEXT := 'pgmq.' || qtable;
+    atable TEXT := pgmq.format_table_name(queue_name, 'a');
+    fq_atable TEXT := 'pgmq.' || atable;
+    partitioned BOOLEAN;
+BEGIN
+    EXECUTE FORMAT(
+        $QUERY$
+        SELECT is_partitioned FROM pgmq.meta WHERE queue_name = %L
+        $QUERY$,
+        queue_name
+    ) INTO partitioned;
+
+    -- check if the queue exists
+    IF NOT EXISTS (
+        SELECT 1
+        FROM information_schema.tables
+        WHERE table_name = qtable and table_schema = 'pgmq'
+    ) THEN
+        RAISE NOTICE 'pgmq queue `%` does not exist', queue_name;
+        RETURN FALSE;
+    END IF;
+
+    IF pgmq._extension_exists('pgmq') THEN
+        EXECUTE FORMAT(
+            $QUERY$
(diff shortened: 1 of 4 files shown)


v0idpwn (approved):

v0idpwn:
It still can fail due to races, but avoiding the potential races is something we can only do if we get a lock, so +1 for the solution.

ChuckHend:
> but avoiding the potential races is something we can only do if we get a lock

Oh ya good point, I didn't think about that. A similiar race condition exists on `pgmq.create()` right, or is this a different race condition?

v0idpwn:
Same concept. Approach would be to get a lock whenever creating or deleting queues.

v0idpwn:
Great idea! I was thinking of advisory locks, but locking `pgmq.meta` should be better. `EXCLUSIVE` should be enough as we can still read it concurrently for metrics.

PR #393, opened 2025-04-17
resolve race condition on pgmq.create

ChuckHend:
Closes https://github.com/tembo-io/pgmq/issues/394

Uses an advisory lock to force pgmq.create() to be executed serially across client processes.

Other options considers:

### 1. Exclusive Row Locks
- check if the queue record exists on pgmq.meta
- get a lock on it if exists, otherwise create it then lock it
- hold the lock and continue on with `create table if not exists` statements as normal

This works, but the implementation is more complex then using the advisory lock.

### 2. Conditionally skip `create table if not exists`
- check if queue record exists on pgmq.meta
- if record exists, skip the rest of pgmq.create. otherwise, create it

I _think_ the race condition still exists. pgmq would not recreate the queue and archive table if the record exists in pgmq.meta, which would be a behavior change.


ChuckHend:
Going with what @v0idpwn suggested [here](https://github.com/tembo-io/pgmq/pull/385#issuecomment-2758807919)

pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
```diff
@@ -102,3 +102,155 @@ BEGIN
...
+CREATE FUNCTION pgmq.create_non_partitioned(queue_name TEXT)
+RETURNS void AS $$
+DECLARE
+  qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+  qtable_seq TEXT := qtable || '_msg_id_seq';
+  atable TEXT := pgmq.format_table_name(queue_name, 'a');
+  lock_key BIGINT;
+BEGIN
+  PERFORM pgmq.validate_queue_name(queue_name);
+  PERFORM pg_advisory_xact_lock(PERFORM pg_advisory_xact_lock(hashtext('pgmq.create_non_partitioned' || queue_name)););
```
v0idpwn, line 116:
```suggestion
  PERFORM pg_advisory_xact_lock(hashtext('pgmq.create_non_partitioned' || queue_name)););
```
typo?

pgmq-extension/sql/pgmq.sql
```diff
@@ -724,9 +724,12 @@ DECLARE
   qtable TEXT := pgmq.format_table_name(queue_name, 'q');
   qtable_seq TEXT := qtable || '_msg_id_seq';
   atable TEXT := pgmq.format_table_name(queue_name, 'a');
+  lock_key BIGINT;
```
v0idpwn, line 727:
```suggestion
```

pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
ChuckHend: Yes

theory (approved):

pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
```diff
@@ -102,3 +102,155 @@ BEGIN
...
+CREATE FUNCTION pgmq.create_non_partitioned(queue_name TEXT)
+RETURNS void AS $$
+DECLARE
+  qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+  qtable_seq TEXT := qtable || '_msg_id_seq';
+  atable TEXT := pgmq.format_table_name(queue_name, 'a');
+  lock_key BIGINT;
+BEGIN
+  PERFORM pgmq.validate_queue_name(queue_name);
+  PERFORM pg_advisory_xact_lock(hashtext('pgmq.create_non_partitioned' || queue_name)););
```
v0idpwn, line 116: Should we get the same lock on drop?

ChuckHend: yeah that is a good idea. i'll push that into this PR too.

v0idpwn (approved):

pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
ChuckHend: pushed that in https://github.com/tembo-io/pgmq/pull/393/commits/b5a0a662c953ecb3f1ddfa5e2018b1c672e4a812

pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
```diff
@@ -253,4 +252,96 @@ BEGIN
...
+RETURNS BOOLEAN AS $$
+DECLARE
+    qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+    qtable_seq TEXT := qtable || '_msg_id_seq';
+    fq_qtable TEXT := 'pgmq.' || qtable;
+    atable TEXT := pgmq.format_table_name(queue_name, 'a');
+    fq_atable TEXT := 'pgmq.' || atable;
+    partitioned BOOLEAN;
+BEGIN
+    PERFORM pg_advisory_xact_lock(hashtext('pgmq.drop_queue' || queue_name));
```
v0idpwn, line 268: I think it makes sense to use the same advisory lock on create and on delete. Then we don't get problems of trying to create and delete the same queue at the same time either. Maybe `hashtext('pgmq_queue_' || queue)`...

ChuckHend: oh ya, great call.

ChuckHend:
@v0idpwn , pelase take another look. Here's summary of changes I think since you approved:

- added `pgmq.acquire_queue_lock()` to use across the project
- call `pgmq.acquire_queue_lock()` in create_non_partitioned, create_unlogged, create_partitioned, drop_queue

v0idpwn (approved):

pgmq-extension/sql/pgmq.sql
```diff
@@ -50,6 +50,16 @@ CREATE TYPE pgmq.queue_record AS (
 -- Functions
 ------------------------------------------------------------
 
+-- prevents race conditions during queue creation by acquiring a transaction-level advisory lock
+-- uses a transaction advisory lock maintain the lock until transaction commit
+-- a race condition would still exist if a session advisory lock is used
```
v0idpwn, line 55:
```suggestion
-- a race condition would still exist if lock was released before commit
```

v0idpwn (approved):

PR #395, opened 2025-04-28
add installation guide

### LINKED ISSUE #388: SQL-only installation tests and docs
PGMQ can be installed in any Postgres as of https://github.com/tembo-io/pgmq/pull/379, for example via 

`psql -f pgmq-extension/sql/pgmq.sql postgres://postgres:postgres@localhost:5432/postgres`

However, this is not documented or tested anywhere yet. Some considerations/thoughts:

1. The extension tests are currently all in a [single file](https://github.com/tembo-io/pgmq/blob/main/pgmq-extension/test/sql/base.sql) and they expect the [create](https://github.com/tembo-io/pgmq/blob/2471e1761a4c24d4748b42a79bd07a6c9dea8ee3/pgmq-extension/test/expected/base.out#L2)/drop extension commands to work. We might need to refactor tests or find a work around.
2. The "extension upgrade test" runs [alter extension update](https://github.com/tembo-io/pgmq/blob/2471e1761a4c24d4748b42a79bd07a6c9dea8ee3/.github/workflows/extension_upgrade.yml#L85) which is built in to Postgres. I'm not completely sure yet we'll parse the migration SQL files for this as SQL-only.
3. For [documentation](https://github.com/tembo-io/pgmq/tree/main/docs), I think we should add a new section dedicated to installation and cover both the extension install method and the SQL only install method, and also briefly touch on when you'd use each method.

ChuckHend:
Add a guide for installation with special focus on installing via the "sql-only" method.

closes #388

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -83,7 +83,7 @@ PGMQ consists of raw SQL objects and can also be installed directly into any Pos
 Simply executing the SQL definition file on your Postgres instance will create all the required objects. This can be accomplished by cloning the repo then running the following commands. This requires `psql` to be installed and available on your `PATH`.
 
 ```bash
-git clone https://github.com/ChuckHend/pgmq.git
+git clone https://github.com/pgmq/pgmq.git
 
 cd pgmq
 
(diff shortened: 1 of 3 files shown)


INSTALLATION.md
```diff
@@ -0,0 +1,120 @@
...
+```bash
+pgxn install pgmq
+```
+
+### From Source
+
+Alternatively, clone the repository:
+
+```bash
+git clone https://github.com/ChuckHend/pgmq.git
```
v0idpwn, line 26:
```suggestion
git clone https://github.com/pgmq/pgmq.git
```

v0idpwn (approved):

ChuckHend:
cc @axelfontaine

PR #400, opened 2025-05-12
Add pgflow.dev to "Who uses pgmq?" section

jumski:
Adding pgflow to the list – it's a Postgres-native workflow engine built for Supabase projects. Workflows are orchestrated via SQL functions that enqueue tasks using pgmq; a dedicated serverless worker then processes the tasks and acknowledges completion back to the SQL core, which can trigger further jobs on the queues. Uses pgmq’s visibility timeout for retries, which works great.

Cheers!


ChuckHend (approved):

ChuckHend:
Thank you!

PR #401, opened 2025-05-23
fix: Override types of timestamp columns to avoid `time` conflict

spencewenski:
Problem
-------
If a consumer has both the `time` and `chrono` features of `sqlx` enabled, `sqlx` will default to using `OffsetDateTime` (from [`time`](https://docs.rs/time)) for timestamp fields instead of `DateTime<Utc>` (from [`chrono`](https://docs.rs/chrono)). It's easy to accidentally encounter this situation, either by directly enabling `time`, e.g., if the consumer prefers `time` over `chrono`, or if the `time` feature is enabled by a transitive dependency. When this situation occurs, the `pg_ext::PGMQueueMeta` struct can not be built from the sqlx query because an `OffsetDateTime` was returned instead of a `DateTime<Utc>`, causing a compile error.

Solution
--------
`sqlx` provides a way to override the type of an output column: https://docs.rs/sqlx/latest/sqlx/macro.query.html#force-a-differentcustom-type. This PR updates the failing queries in the `pg_ext` to use this approach. Unfortunately, I think this requires using the more verbose approach of naming each column to be returned instead of using `*`.

Testing
-------
- I ran `cargo test` locally and all the tests passed
- I ran the examples and they appeared to run successfully
- I enabled the `time` feature of `sqlx` in the `dev-dependencies` to ensure this case of enabling `chrono` and `time` is supported, which can be confirmed by compiling examples/tests.

Other
-----
I also updated the sqlx query caches by running `cargo sqlx prepare` locally. I noticed there was an unused column called `headers` in some of the queries that used a `*` selector. I can add the column back, but it didn't appear to be used, so I left it out.

### DIFF
--- a/pgmq-rs/Cargo.toml
+++ b/pgmq-rs/Cargo.toml
@@ -12,10 +12,10 @@ readme = "README.md"
 repository = "https://github.com/tembo-io/pgmq"
 
 [dependencies]
-chrono = { version = "0.4.23", features = [ "serde" ] }
+chrono = { version = "0.4.23", features = ["serde"] }
 serde = { version = "1.0.152" }
-serde_json = { version = "1.0.91", features = [ "raw_value" ] }
-sqlx = { version = "0.8.1", features = [ "runtime-tokio" , "postgres", "chrono", "json" ] }
+serde_json = { version = "1.0.91", features = ["raw_value"] }
+sqlx = { version = "0.8.1", features = ["runtime-tokio", "postgres", "chrono", "json"] }
 thiserror = "1.0.38"
 tokio = { version = "1", features = ["macros"] }
 log = "0.4.17"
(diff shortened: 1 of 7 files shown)


pgmq-rs/src/pg_ext.rs
```diff
@@ -183,7 +183,7 @@ impl PGMQueueExt {
         &self,
         executor: E,
     ) -> Result<Option<Vec<PGMQueueMeta>>, PgmqError> {
-        let queues = sqlx::query!("SELECT * from pgmq.list_queues();")
+        let queues = sqlx::query!(r#"SELECT queue_name, is_partitioned, is_unlogged, created_at as "created_at: chrono::DateTime<Utc>" from pgmq.list_queues();"#)
```
spencewenski, line 186: I wasn't sure if you have a preferred style for long query lines. I'm happy to adjust the formatting if you'd like.

pgmq-rs/Cargo.toml
```diff
@@ -29,3 +29,7 @@ env_logger = "0.10.0"
 rand = "0.8.5"
 regex = "1.5.4"
 lazy_static = "1.4.0"
+# Enable the `time` feature in dev dependencies to ensure we are able to support consumers who have
+# the feature enabled, either directly or via a transitive dependency. See the following issue for
+# more details: https://github.com/launchbadge/sqlx/issues/3412
+sqlx = { version = "0.8.1", features = ["time"] }
```
ChuckHend, line 35: I think your code changes to the SQL queries might solve the issue, and this line is not required?

I checked out your code, then added "time" to the sqlx features as shown below and and able to compile successfully and run tests. Maybe I missed something though.

```toml
sqlx = { version = "0.8.1", features = [ "runtime-tokio" , "postgres", "chrono", "json", "time" ] }
```

pgmq-rs/src/pg_ext.rs
ChuckHend: I kind of like this better, having the names explicitly stated.

The unused `headers` field is a feature that has been added to the database extension but hasn't made it into the Rust client quite yet.

pgmq-rs/Cargo.toml
spencewenski: Correct, but if you remove my code changes, but keep this change, then things won’t compile. I added this to the dev dependencies to catch any new queries that would see the conflict with the time crate.

spencewenski: I don’t recommend adding the time feature in the main dependency section, as that will affect consumers. Adding it here allows you to only enable it for tests.

PR #412, opened 2025-06-30
detach and attach queues

### LINKED ISSUE #408: PGMQ Queue Data Not Restoring from Backup
I've encountered an issue when trying to back up and restore PGMQ queue data
```
# Full database backup
pg_dump -U <user> -d <db> -F p -f db.dump
# PGMQ schema-specific backup
pg_dump -U <user> -d <db> -n pgmq --file=pgmq_backup.sql
```
I drop and re-create DB
After restoring the backup, running SELECT * FROM pgmq.list_queues(); returns empty results

Does pg_dump fully support backing up PGMQ queue tables ?
Are there special considerations or additional steps required to back up PGMQ queues ?

ChuckHend:
Proposed solution to making queue data compatible with `pg_dump`. The process for this would be:

1. Remove the queue objects from the extension's membership.

```sql
select pgmq.detach_all()
```
2. Dump the extension schema:

```bash
pg_dump --dbname=postgres://postgres:postgres@localhost:5432/postgres --schema pgmq --extension pgmq > pgmq-dump.sql
```

3. Restore into new postgres instance (with pgmq extension already installed, but not yet ran with `create extension pgmq`
```bash
psql --dbname=postgres://postgres:postgres@localhost:5432/postgres < pgmq-dump.sql
```

4. Re assign the objects back to the extension
```sql
select pgmq.attach_all();
```

Then continue on using the extension as normal.

Closes #408

### DIFF
--- a/pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
+++ b/pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
@@ -435,3 +435,117 @@ BEGIN
 
 END;
 $$ LANGUAGE plpgsql;
+
+-- detach all pgmq objects
+-- note--this will prevent `DROP EXTENSION pgmq CASCADE` from being able to drop the extension objects
+CREATE OR REPLACE FUNCTION pgmq."detach_all"()
+RETURNS TABLE(detached_object TEXT)
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    queue_record RECORD;
+    detach_cmd TEXT;
+BEGIN
+    -- Only proceed if pgmq extension exists (not applicable when SQL-only)
+    IF NOT pgmq._extension_exists('pgmq') THEN
+        RAISE NOTICE 'pgmq extension does not exist, nothing to detach';
+        RETURN;
+    END IF;
+
+    -- Iterate through all queues and detach their objects (table, archive table, sequence)
+    FOR queue_record IN SELECT queue_name FROM pgmq.meta LOOP
+        BEGIN
+            detach_cmd := format('ALTER EXTENSION pgmq DROP TABLE pgmq.q_%I', queue_record.queue_name);
+            EXECUTE detach_cmd;
+            detached_object := format('TABLE pgmq.q_%s', queue_record.queue_name);
+            RETURN NEXT;
+        EXCEPTION WHEN OTHERS THEN
+            RAISE WARNING 'Failed to detach queue table for %: %', queue_record.queue_name, SQLERRM;
+        END;
+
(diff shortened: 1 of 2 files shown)


ChuckHend:
@theory , @v0idpwn , curious your thoughts on this as it relates to #408

v0idpwn (commented):
The concept looks great to me. Curious about @theory's opinion on it.

pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
```diff
@@ -435,3 +435,117 @@ BEGIN
...
+RETURNS TABLE(detached_object TEXT)
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    queue_record RECORD;
+    detach_cmd TEXT;
+BEGIN
+    -- Only proceed if pgmq extension exists (not applicable when SQL-only)
+    IF NOT pgmq._extension_exists('pgmq') THEN
+        RAISE NOTICE 'pgmq extension does not exist, nothing to detach';
```
theory, line 451: Nit: maybe say "not installed", since it obviously _exists_.

pgmq-extension/sql/pgmq--1.5.1--1.5.2.sql
```diff
@@ -435,3 +435,117 @@ BEGIN
...
+        RETURN;
+    END IF;
+
+    -- Iterate through all queues and detach their objects (table, archive table, sequence)
+    FOR queue_record IN SELECT queue_name FROM pgmq.meta LOOP
+        BEGIN
+            detach_cmd := format('ALTER EXTENSION pgmq DROP TABLE pgmq.q_%I', queue_record.queue_name);
```
theory, line 458:
```suggestion
            detach_cmd := format('ALTER EXTENSION pgmq DROP TABLE pgmq.%I', 'q_'  || queue_record.queue_name);
```

theory: This change needs to be made throughout, or else table names will render as `q_"Foo"` instead of the expected `"q_Foo"`.

theory (approved):
I think this is a fine workaround, but honestly this feels like a bit of design issue, like it shouldn't be necessary. Would it be possible to support queue objects outside of the extension? I think that would represent a better separation of concerns: Queuing logic vs. storage.

ChuckHend:
Closing this PR in favor of https://github.com/pgmq/pgmq/pull/413. We'll change the membership model such that the queue table and a sequence, and archive table are not members of the extension.

PR #432, opened 2025-09-06
Add new queue notification feature

chrisprobst:
Add new queue notification feature that notifies listeners about inserts, updates and deletions of messages. This is opt-in and can be enabled/disabled on a per-queue basis. This feature requires client support as LISTEN/NOTIFY alone do not help much. This is a starting point to allow real-time updates additionally to polling.

### DIFF
--- a/pgmq-extension/sql/pgmq.sql
+++ b/pgmq-extension/sql/pgmq.sql
@@ -1134,29 +1134,29 @@ BEGIN
 END;
 $$ LANGUAGE plpgsql;
 
-CREATE OR REPLACE FUNCTION pgmq.enable_notifications(queue_name TEXT)
+CREATE OR REPLACE FUNCTION pgmq.enable_notify_insert(queue_name TEXT)
 RETURNS void AS $$
 BEGIN
   EXECUTE FORMAT(
     $QUERY$
-    CREATE CONSTRAINT TRIGGER trigger_notify_queue_listeners
-    AFTER INSERT OR UPDATE OR DELETE ON pgmq.%I
+    CREATE CONSTRAINT TRIGGER trigger_notify_queue_insert_listeners
+    AFTER INSERT ON pgmq.%I
     DEFERRABLE FOR EACH ROW
     EXECUTE PROCEDURE pgmq.notify_queue_listeners()
     $QUERY$,
-    'q_' || queue_name
+    ('q_' || queue_name)::regclass
   );
 END;
 $$ LANGUAGE plpgsql;
 
-CREATE OR REPLACE FUNCTION pgmq.disable_notifications(queue_name TEXT)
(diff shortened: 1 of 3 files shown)


pgmq-extension/sql/pgmq.sql
```diff
@@ -1125,3 +1125,39 @@ BEGIN
...
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE OR REPLACE FUNCTION pgmq.enable_notifications(queue_name TEXT)
+RETURNS void AS $$
+BEGIN
+  EXECUTE FORMAT(
+    $QUERY$
+    CREATE CONSTRAINT TRIGGER trigger_notify_queue_listeners
+    AFTER INSERT OR UPDATE OR DELETE ON pgmq.%I
```
ChuckHend, line 1143: We might only want it on `INSERT`?

`UPDATE` will happen when a message read from the queue (vt and read_ct are updated).

ChuckHend: Or do you think we could have client listen for each OP seprately?

```sql
LISTEN pgmq.q_my_queue.INSERT;
LISTEN pgmq.q_my_queue.UPDATE;
LISTEN pgmq.q_my_queue.DELETE;
```

chrisprobst: Yes I thought maybe there is a use case for that. I can live with INSERT-only as well. I will continue on Monday.

ChuckHend:
I will figure out what is going on with the failing upgrade test. I don't think it is related to this PR.

pgmq-extension/sql/pgmq.sql
chrisprobst: I thought about this again. So, yes, I think selective notifications are valuable to have, even thought I agree and INSERT should be probably the most interesting one. I am also fine with going for INSERT only initially. We can go from there. If you agree, we can rename to enable_notify_insert(queue_name TEXT) to make the effect clear.

ChuckHend (approved):

chrisprobst:
@ChuckHend Thank you very much! Did 1. and 2., let me know your thoughts on testing.

Regarding idempotency, we could do:

DROP TRIGGER IF EXISTS ...

As part of the enable_ function and also add the IF EXISTS clause to the disable function which would allow calling it over and over again.

What do you think?

ChuckHend:
@chrisprobst, let's go with what you suggested and add "DROP TRIGGER IF EXISTS" to both the enable and disable functions.

For testing, I have a [draft](https://github.com/pgmq/pgmq/pull/434) up. I'll rebase that branch once this PR merges, then ask for your review on those tests.

chrisprobst:
@ChuckHend Ok, I simply call disable_ from enable_ now to not replicate the logic. Should be fine because functions are always part of the current transaction. The latest sate is pushed.

PR #433, opened 2025-09-08
added pop many and added migration

legion49f:
If there is anything else or a better way to do this please let me know


pgmq-extension/sql/pgmq.sql
```diff
@@ -536,8 +536,8 @@ BEGIN
 END
 $$ LANGUAGE plpgsql;
 
--- pop a single message
-CREATE FUNCTION pgmq.pop(queue_name TEXT)
+-- pop: actual implementation
+CREATE FUNCTION pgmq.pop(queue_name TEXT, qty INTEGER)
```
v0idpwn, line 540: I think for simplicity you can use a default in this argument. We use the overloading pattern in some cases but they are more complex.

legion49f: @v0idpwn changed to use `DEFAULT`

v0idpwn (approved):
Looks great, can we just add a test?

v0idpwn:
SQL tests are better :)
They work by diffing the output. You add your query on `base.sql` and the query + expected output on base.out. It will then compare both and if it's different, it fails.

v0idpwn:
You can use the makefile to install pgmq into your postgres and run tests against it.

v0idpwn:
Some docs about this here: https://www.postgresql.org/docs/current/extend-pgxs.html

pgmq-extension/test/expected/base.out
```diff
@@ -577,6 +577,39 @@ SELECT msg_id = :first_msg_id FROM pgmq.pop('test_pop_queue');
...
+    6
+(1 row)
+
+SELECT * from pgmq.pop('test_pop_queue', 3);
+ msg_id | read_ct |          enqueued_at          |              vt               | message | headers
+--------+---------+-------------------------------+-------------------------------+---------+---------
+      4 |       0 | 2025-09-09 20:03:57.971415+00 | 2025-09-09 20:03:57.97151+00  | 1       |
+      6 |       0 | 2025-09-09 20:03:57.975432+00 | 2025-09-09 20:03:57.975481+00 | 3       |
+      5 |       0 | 2025-09-09 20:03:57.973577+00 | 2025-09-09 20:03:57.973647+00 | 2       |
+(3 rows)
```
v0idpwn, line 611: These dates aren't gonna match, as they are different on every run. You can select only `msg_id`,

ChuckHend (approved):
Awesome. Thank you!

pgmq-extension/test/expected/base.out
legion49f: I have setup my own runners at home so i can get this test to pass. Ill update when its ready

legion49f:
image
Looks like i finally got the sql output to match. Thank you for the patience!

pgmq-extension/sql/pgmq--1.6.1--1.6.2.sql
```diff
@@ -0,0 +1,31 @@
+-- Upgrade pgmq from 1.6.1 to 1.6.2
```
legion49f, line 1: Should I delete this file since you guys are almost at 1.7.0?

legion49f:
Running the tests was really painful. Is it okay if I add a make target to do exactly the same thing as the github runners do but with a volume mount on docker so that the resultant base.out gets save locally on the users laptop? (different PR of cource :) )

pgmq-extension/sql/pgmq--1.6.1--1.6.2.sql
ChuckHend: Yes

legion49f:
I saw a draft PR about tests failing intermittently. Are these test failures to do with my PR?

ChuckHend:
> I saw a draft PR about tests failing intermittently. Are these test failures to do with my PR?

No not related to this PR, but I think the upgrade test is failing here because of that flake. I might have a fix, I'll ge that merged into main quick so you can pull that in. Hopefully it resolves the failing upgrade test on this PR.

legion49f:
@ChuckHend, @v0idpwn  are we good to merge?

PR #434, opened 2025-09-09
Client-side notification tests

ChuckHend:
Tests for notification feature implemented in https://github.com/pgmq/pgmq/pull/432. Wait for that to merge

### DIFF
--- a/.github/workflows/client-tests.yml
+++ b/.github/workflows/client-tests.yml
@@ -15,6 +15,7 @@ on:
   push:
     branches:
       - main
+      - client-tests
     paths:
       - "pgmq-extension/**"
       - ".github/workflows/client-tests.yml"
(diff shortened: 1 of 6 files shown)


ChuckHend:
@chrisprobst, could you take a look through these tests and let me know what you think?

.github/workflows/extension_upgrade.yml
```diff
@@ -66,6 +66,7 @@ jobs:
           cargo install cargo-pgrx --version 0.11.0
           SQLX_OFFLINE=true cargo pgrx run pg16 --pgcli || true
           export DATABASE_URL=postgres://$USER:postgres@localhost:28816/pgmq
+          psql $DATABASE_URL -c "DROP SCHEMA IF EXISTS pgmq CASCADE;"
```
ChuckHend, line 69: @v0idpwn , im sneaking a patch in for the extension upgrade tests. It is really hard to reproduce, but I am pretty sure the issue with failing upgrade was that the "SQL only" tests were leaving objects lingering around. `DROP EXTENSION` wouldnt clear out those objects, but `DROP SCHEMA` does.

ChuckHend: I merged this in https://github.com/pgmq/pgmq/pull/436 instead.

v0idpwn (approved):

PR #436, opened 2025-09-10
Fix failing upgrade test

ChuckHend:
Seems like the SQL-only tests are colliding with these upgrade tests. It has been really hard to reproduce, but I think this is resolving it.


v0idpwn:
Makes a lot of sense!

PR #448, opened 2025-11-03
Sync documentation with source + add sync command

jumski:
I went through all the SQL API documentation and synced it with the actual source code in `pgmq.sql`. Found and fixed 29 issues including missing function overloads, incorrect parameter types, missing type definitions, and outdated examples. Also fixed the README examples to show correct output and default values.

I've added a Claude command at `.claude/commands/sync-docs-and-readme.md` that can be used to repeat this process in the future. It works as `/sync-docs-and-readme` in Claude Code, but the guide is AI-agnostic so you can copy it to ChatGPT or any other AI assistant if needed. Ran the command after making all changes to verify everything is now in sync.

See individual commit messages for detailed lists of all changes.

---

## What Changed

**Types Documentation** (`docs/api/sql/types.md`)

- Added missing `headers` field to `message_record`
- Fixed `read_ct` type (bigint → integer)
- Added `queue_record` and `metrics_result` type docs

**Functions Documentation** (`docs/api/sql/functions.md`)

- Documented all 6 `send()` overloads (was 1)
- Documented all 6 `send_batch()` overloads (was 1)
- Added 4 missing functions: `create_non_partitioned()`, `convert_archive_partitioned()`, `enable_notify_insert()`, `disable_notify_insert()`
- Fixed parameter names, types, and defaults across all functions
- Added missing `queue_visible_length` field to metrics
- Documented queue name limit (47 chars) and pg_partman requirement

**README** (`pgmq-extension/README.md`)

- Added missing `headers` column to all examples
- Fixed `create_partitioned()` defaults: `'daily'` → `'10000'`, `'5 days'` → `'100000'`

All function signatures, types, defaults, and examples now match the actual implementation in `pgmq.sql`.


ChuckHend:
hey thank you @jumski! Nice idea to keep the documentation updated, I like it. Maybe we could make that command `/pgmq-update-docs`? Just a bit more concise, and the prefix with `pgmq` will help ensure it doesn't collide with any other claude commands folks have defined in `~/.claude/commands`.

jumski:
@ChuckHend it is already generic and you should be able to just paste it into Cursor or other agent - it does not contain any specific claude code formatting or tool names (intentional).

The only non-generic thing about this command is that it is placed in claude-specific location, but we can also fix that by symlinking it to something like `SYNC_DOCS_WITH_AI.md` or similar (i'm only using claude so i dont know what other tools are using).

PR #459, opened 2025-11-29
add notify and throttle doc

ChuckHend:
@smyrgeorge , @chrisprobst can you take a look at what I have written in these docs and make some more suggestions if necessary?

Once this merges I will work on getting v1.8.0 released and close https://github.com/pgmq/pgmq/issues/431


chrisprobst:
LGTM from my side. Thanks.

PR #479, opened 2026-01-13
Add another Haskell PGMQ client library (stakhanov) to Client Libraries

MichelBoucey:
BTW : I am the author of this one.


ChuckHend:
Thanks @MichelBoucey. Is this an alternative to the other Haskell client, different from the other one in some way?

PR #486, opened 2026-01-31
Support for AMQP like topics

smyrgeorge:
Please accept the PR that adds AMQP style topics to pgmq.

### How it works
The implementation It's build around the `send` function.
You send a message to the `send_topic` and the message will be delivered to 0..N queues by matching the routing key.
No any other modifications added to the rest of the code.

See here also: issue https://github.com/pgmq/pgmq/issues/455

If you are OK with the implementation (design and API) I can proceed and write some tests also

### DIFF
--- a/pgmq-extension/client-tests/topics_test.py
+++ b/pgmq-extension/client-tests/topics_test.py
@@ -1076,5 +1076,180 @@ class TestSendTopic:
             drop_queue(db_connection, queue_api_errors)
 
 
+# =============================================================================
+# Tests for regex injection protection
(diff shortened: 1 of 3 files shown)


ChuckHend:
This looks good @smyrgeorge. Do you think there will be additional kinds of exchanges implemented over time?

smyrgeorge:
@ChuckHend 
I just add the `topics_test.py`, topics.md` and `pgmq--1.10.0--1.11.0.sql` files.
Any other comments/ideas are welcome. Let me know. 😄

PR #500, opened 2026-02-16
prep release v1.11.0

ChuckHend:


ChuckHend:
bumping control version over in https://github.com/pgmq/pgmq/pull/501 instead

PR #503, opened 2026-02-19
Update types.md

paulovieira:
Show an updated (and full) example of `message_record`


ChuckHend:
Thank you @paulovieira

PR #506, opened 2026-02-24
Add Elixir pgflow implementation to README.md

themusicman:


ChuckHend:
Hey @themusicman . https://github.com/agoodway/pgflow is a good project and I would like to put it on the readme. I think it is distinct from a client library though (pgflow is a lot more than that). Can we move it down under https://github.com/pgmq/pgmq?tab=readme-ov-file#who-uses-pgmq? We could rename that to "Projected Powered by PGMQ" or something better. @cpursley what do you think?

PR #508, opened 2026-02-26
feat!: Improve sql-only installation from the Rust client

### LINKED ISSUE #417: create SQL-only upgrade tests
The extension tests run `alter extension pgmq update` in order to test the migration SQL is valid. However, this does not properly test the migration files for the "sql-only" installation methods.

### LINKED ISSUE #504: More robust installation from the Rust client
Hello, I'm interested in installing the extension via the Rust client because I will likely be using a managed postgres provider that doesn't (yet) support the pgmq extension. I saw [this previous issue](https://github.com/pgmq/pgmq/issues/439); however, the current approach does not seem reliable enough for my needs. Specifically:

- I don't want to fetch the installation script from GitHub. This will introduce instability into my service (I would plan to run the installation as part of the app's boot process, similar to how I run my diesel migrations).
- I would want to ensure that the pgmq client version I'm using is compatible with the pgmq extension version that gets installed
- I would want pgmq extension upgrades to be applied automatically (vs, the current approach only runs the main pgmq.sql script)

Some possible improvements were mentioned in [the previous issue](https://github.com/pgmq/pgmq/issues/439); to summarize, I think the approach would look something like the following:

- Enable installing the pgmq extension via a list of migration files 
- Provide the migration files as part of the pgmq rust crate (for example, by embedding them using [include_dir](https://crates.io/crates/include_dir))
- Add a method to the rust client to run the migration files

I see that there are already a set of migration scripts; would we be able to re-use those for this purpose? If so, I think there are a couple additional considerations:

- I see that the first migration file starts at version 0.7.3 (and it's actually empty), so I assume we would to add some "initial setup" script that installs the 0.7.3 version of pgmq? Then, running the the other migrations would (I assume) update pgmq to the latest version?
- I assume the migration scripts that currently exist are not meant to run more than once, so there would need to be a way to remember which scripts were already run, or at minimum which version of pgmq is currently installed. Diesel, for example, solves this by storing which migrations were already run in a table it manages.

Is this something you'd be open to adding to the pgmq crate? Let me know if there's anything else I'm missing or if there's another approach you'd prefer.

spencewenski:
## Problem
As described in https://github.com/pgmq/pgmq/issues/504, the current SQL-only installation method is not ideal for production use because

1. It fetches the installation script from GitHub, which can introduce instability into a service that runs the installation during boot along with its other migrations.
2. There's no guarantee that the installed version is compatible with the current version of the `pgmq` client.
3. The installation is not versioned, making it difficult to upgrade `pgmq` to a new version.
4. There is no locking mechanism, so if multiple processes attempt to install `pgmq` at the same time, they may conflict and fail to install

## Solution
This PR solves the above problems for the Rust client. It does so by:

1. Adding a new installation method that uses SQL scripts that are embedded into the `pgmq` crate, eliminating the need to download the scripts from GitHub at runtime. This also means that the Rust client will always install the latest version of `pgmq` that it's compatible with.
2. Tracking which migration scripts were run in a table managed by the Rust client, along with the version associated with each script. This allows upgrading `pgmq` by running all migrations greater than the last run migration.
3. Using locks to ensure that only a single process can run the SQL-only installation at once. This approach will work across different client implementations as long as they use the same locks (one advisory lock and one table-level lock).

All of the above is implemented with pure-PostgreSQL functionality, so it should be easy to implement in other clients.

This PR also:
1. Updates the `crate_ci.yml` and `extension_upgrade.yml` workflows to ensure the SQL-only installation is tested just as thoroughly as the extension installation.
2. Adds a utility method to initialize the "applied migrations" table to enable switching from the previous unversioned SQL-only installation method to the new versioned method(s). This is required for the use case described in https://github.com/pgmq/pgmq/issues/514)
3. Adds a utility method to fetch the currently installed version of PGMQ (assuming it was installed with the new versioned SQL-only method).

Closes https://github.com/pgmq/pgmq/issues/504
Closes https://github.com/pgmq/pgmq/issues/417


pgmq-rs/Cargo.toml
```diff
@@ -18,7 +18,8 @@ required-features = ["cli"]
 
 [features]
 default = []
-cli = ["clap", "futures-util", "reqwest", "tokio/rt"]
+cli = ["clap", "url/serde", "install-sql"]
+install-sql = ["futures-util", "include_dir", "regex", "itertools"]
```
ChuckHend: might as well use or rename the `cli` feature. the only thing the cli currently does is install the SQL.

pgmq-rs/src/bin/pgmq-cli.rs
```diff
@@ -1,50 +1,43 @@
...
                 .expect("Failed to connect to database");
 
-            if let Err(e) = install_pgmq(&pool, version).await {
-                log::error!("Error installing PGMQ: {}", e);
-                process::exit(1);
-            }
-        }
-        _ => {
-            log::error!("No valid subcommand provided. Use --help for usage information.");
-            process::exit(1);
```
spencewenski: Clap handles this case automatically when using derives.

pgmq-rs/src/install/version.rs
```diff
@@ -0,0 +1,202 @@
...
+/// The `pgmq` extension control file. Used to determine which version of `pgmq` would be
+/// installed by running the `pgmq.sql` script.
+static EXTENSION_CONFIG: &str = include_str!(concat!(
+    env!("CARGO_MANIFEST_DIR"),
+    "/../pgmq-extension/pgmq.control"
+));
+
+/// Struct to represent a basic semver version, e.g. `1.2.3`.
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct Version {
```
spencewenski: We could potentially use `https://crates.io/crates/semver` instead of creating our own `Version` struct, but the version format we need to support is more limited than what's supported by `semver`, so I opted to create this instead.

pgmq-rs/Cargo.toml
ChuckHend: Ok that makes sense. I agree.

pgmq-rs/src/install/applied.rs
```diff
@@ -72,8 +72,8 @@ impl AppliedMigration {
             .into_iter()
             .map(|row| -> Result<AppliedMigration, PgmqError> {
                 Ok(Self {
-                    name: row.get::<String, _>("name"),
-                    version: Version::from_str(&row.get::<String, _>("version"))?,
+                    name: row.try_get::<String, _>("name")?,
+                    version: Version::from_str(&row.try_get::<String, _>("version")?)?,
```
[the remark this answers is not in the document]
spencewenski: I tried that briefly but got a little stuck because I needed to implement a few traits for `Version`. I can try again, though, and worst case create another private struct that uses String for the version field and then convert to the public struct.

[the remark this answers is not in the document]
spencewenski: Updated to use `query_as`. Let me know if you know of a better way to implement `Decode`/`Encode`/`Type` for `Version` -- it didn't seem possible to simply derive `Type` from what I could find.

INSTALLATION.md
```diff
@@ -92,7 +96,28 @@ psql -f pgmq-extension/sql/pgmq.sql postgres://postgres:postgres@localhost:5432/
 
 Replace the postgres user, password, and database with the appropriate values for your Postgres instance.
 
-## Uninstalling the SQL-only install
+### Versioned installation
+
+Some clients provide a versioned SQL-only installation. This allows either a fresh installation or `pgmq`, or upgrading
```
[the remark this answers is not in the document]
spencewenski: Oops, will update

pgmq-rs/src/install/script.rs
```diff
@@ -0,0 +1,389 @@
...
+use std::borrow::Cow;
+use std::cmp::Ordering;
+use std::str::FromStr;
+use std::sync::OnceLock;
+
+/// The name of the migration script used to perform a fresh installation of `pgmq`.
+static INIT_SCRIPT_NAME: &str = "pgmq.sql";
+
+/// All of the extension's migration scripts.
+static MIGRATION_SCRIPTS: Dir<'static> = include_dir!("$CARGO_MANIFEST_DIR/../pgmq-extension/sql/");
```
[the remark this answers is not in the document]
spencewenski: Hmm, I thought this would work since it compiles locally and in CI but maybe not. I will double check.

Same comment for the `pgmq.control` file.

pgmq-extension/client-tests/upgrade_test.py
```diff
@@ -182,10 +182,22 @@ def purge_queue(conn, queue_name):
...
+        version = row[0] if row else None
+
+    if version is None:
+        import semver
+        with conn.cursor() as cur:
+            cur.execute("SELECT version FROM pgmq.__pgmq_migrations")
+            rows = cur.fetchall()
+            rows = [semver.Version.parse(row[0]) for row in rows]
+            rows.sort()
+            version = rows[0] if len(rows) > 0 else None
```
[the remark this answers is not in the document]
ChuckHend: Makes sense I think for this function to get the current version. Or change the name of the helper.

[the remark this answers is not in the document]
spencewenski: Yes, I fixed this already.

pgmq-rs/src/install/pgmq.control
```diff
@@ -0,0 +1 @@
+../../../pgmq-extension/pgmq.control
```
ChuckHend: Is this symlinked here so that it is included in the build, or serving a different purpose?

ChuckHend (commented):
Sorry for delay @spencewenski, there is a lot here. Are there any tests covering installation and upgrading to specific versions? I'm not familiar with the `insta` project, so maybe its in there and I am not understanding it. The current API allows the user to pass the version they want to install, but it seems like that got dropped.

pgmq-rs/src/install/pgmq.control
spencewenski: Yes, as far as I can tell, this is the only way to get files from outside the crate's root directory (the directory that contains the `Cargo.toml` file) to be included in the crate's public build.

ChuckHend: Could you use a [build script](https://doc.rust-lang.org/cargo/reference/build-scripts.html) for this? Read from the SQL dir then write to OUT_DIR?

ChuckHend (approved):

PR #518, opened 2026-03-31
chore: Add docs check and minimal versions check

spencewenski:
This PR adds the following checks to the crate's `Makefile` and CI:

1. `check.docs` -- Checks that the docs build without any errors/warnings.
2. `check.minimal-versions` -- Checks that our dependencies are declared as the correct minimal versions. For example, if we depend on a crate with version "1.0.0", ensure we don’t accidentally depend on a feature added in version "1.5.0".

This PR also fixes the errors raised by these checks.

### DIFF
--- a/pgmq-rs/src/pg_ext/mod.rs
+++ b/pgmq-rs/src/pg_ext/mod.rs
@@ -46,7 +46,7 @@ impl PGMQueueExt {
     #[cfg(feature = "install-sql-github")]
     #[deprecated(
         note = "Use install_sql_from_github_with_cxn/install_sql_from_github or install_sql_embedded_with_cxn/install_sql_embedded instead.",
-        since = "0.32.2"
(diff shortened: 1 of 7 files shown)


pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -46,7 +46,7 @@ impl PGMQueueExt {
     #[cfg(feature = "install-sql-github")]
     #[deprecated(
         note = "Use install_sql_from_github_with_cxn/install_sql_from_github or install_sql_embedded_with_cxn/install_sql_embedded instead.",
-        since = "0.32.2"
+        since = "0.33.0"
```
spencewenski: I meant to update this to `0.33.0` when I updated the crate version in https://github.com/pgmq/pgmq/pull/508

ChuckHend (approved):

PR #523, opened 2026-04-07
feat: Add "send batch" functionality to `PGMQueueExt`

spencewenski:
This PR adds support for batch enqueueing messages from the `PGMQueueExt` rust client.

This PR also updates the `make check.clippy` command to disallow clippy warnings instead of just errors. This will cause clippy to emit an error for unused imports and any other warning-level lint.

Relates to https://github.com/pgmq/pgmq/issues/494

### DIFF
--- a/pgmq-rs/src/pg_ext/mod.rs
+++ b/pgmq-rs/src/pg_ext/mod.rs
@@ -390,15 +390,15 @@ impl PGMQueueExt {
         check_input(queue_name)?;
         let delay: VisibilityTimeoutOffset = delay.into();
         let msg = serde_json::to_value(message)?;
-        let sent = sqlx::query(
-            "SELECT send as msg_id from pgmq.send(queue_name=>$1::text, msg=>$2::jsonb, delay=>$3::int);",
+        let msg_id: i64 = sqlx::query_scalar(
+            "SELECT * from pgmq.send(queue_name=>$1::text, msg=>$2::jsonb, delay=>$3::int);",
         )
-            .bind(queue_name)
-            .bind(msg)
-            .bind(delay)
-            .fetch_one(executor)
-            .await?;
-        Ok(sent.try_get("msg_id")?)
+        .bind(queue_name)
+        .bind(msg)
+        .bind(delay)
+        .fetch_one(executor)
+        .await?;
+        Ok(msg_id)
     }
 
     pub async fn send_delay<T: Serialize>(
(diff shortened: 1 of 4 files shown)


pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -396,15 +389,15 @@ impl PGMQueueExt {
     ) -> Result<i64, PgmqError> {
         check_input(queue_name)?;
         let delay: VisibilityTimeoutOffset = delay.into();
-        let msg = serde_json::json!(&message);
```
spencewenski, line 399: I switched away from `serde_json::json!` because it expands to `serde_json::to_value(...).unwrap()`, and it's not good practice to panic in production.

pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -418,6 +411,71 @@ impl PGMQueueExt {
...
+        )
+            .bind(queue_name)
+            .bind(msgs)
+            .bind(delay)
+            .fetch_all(executor)
+            .await?;
+        let sent = sent
+            .into_iter()
+            .map(|row| row.try_get("msg_id"))
+            .collect::<Result<Vec<i64>, _>>()?;
```
ChuckHend: Can we use query_as here and deserialize directly to Vec<i64> or do we have to do the .map() separately?

ChuckHend (approved):
Thanks! I left one comment but it doesn't need to block anything.

pgmq-rs/src/pg_ext/mod.rs
spencewenski: Not sure, I will try that.

spencewenski: `query_as` requires creating a struct to deserialize named columns, or using a tuple. However, a tuple requires at least two elements, so we would need to do something like this:

```rust
let msg_id: (i64, _) = sqlx::query_as(
    "SELECT send as msg_id, 1 from pgmq.send(queue_name=>$1::text, msg=>$2::jsonb, delay=>$3::int);",
)
```

Since there's only one column in the SQL response, we can simply use [ `query_scalar`](https://docs.rs/sqlx/latest/sqlx/macro.query_scalar.html) instead.

ChuckHend: awesome!

PR #524, opened 2026-04-09
feat: Add "read batch" functionality to `PGMQueueExt`

spencewenski:
This PR adds support for reading batches of messages from `PGMQueueExt` rust client. This PR also updates the existing single-message read functions to use the batch method with a `qty` of one.

Relates to https://github.com/pgmq/pgmq/issues/494

### DIFF
--- a/pgmq-rs/src/pg_ext/mod.rs
+++ b/pgmq-rs/src/pg_ext/mod.rs
@@ -5,6 +5,7 @@ use crate::types::{Message, QUEUE_PREFIX};
 use crate::util::{check_input, connect};
 use log::info;
 use serde::{Deserialize, Serialize};
+use sqlx::postgres::PgRow;
 use sqlx::types::chrono::Utc;
 use sqlx::{Acquire, FromRow, Pool, Postgres, Row};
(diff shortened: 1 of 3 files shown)


pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -544,20 +566,9 @@ impl PGMQueueExt {
         .bind(poll_timeout_s)
         .bind(poll_interval_ms)
         .fetch_all(executor)
-        .await;
-
-        match result {
-            Err(sqlx::error::Error::RowNotFound) => Ok(None),
```
spencewenski, line 550: I confirmed that this error is not returned for the `pgmq.read` queries - an empty vec is returned instead when using `fetch_all`. I assumed it’s also not returned for `pgmq.read_with_poll`, but I will confirm.

ChuckHend (approved):

pgmq-rs/src/pg_ext/mod.rs
spencewenski: I added a test to confirm this.

PR #525, opened 2026-04-10
chore: Check that upgrade scripts do not diverge from a fresh install

spencewenski:
## Problem
As seen in https://github.com/pgmq/pgmq/issues/521, it's easy for the upgrade scripts to diverge from a fresh installation.

## Solution
This PR adds a check to verify that the upgrade scripts do not diverge from a fresh installation. The check uses the output of `pg_dump` as well as a manual query of the extensions functions (these are not included in `pg_dump`).

This PR also adds a new migration script (`pgmq--1.11.0--1.11.1.sql`) to resolve the existing differences between the upgrade scripts and a fresh installation.

#### `pgmq.read_grouped_head`

This function only existed in the fresh install (`pgmq.sql`), so this PR adds it via a migration file as well.

#### `pgmq._ensure_pg_partman_installed`
This method exists for both fresh installs and upgrades, but the implementations are different. This PR updates the version in the upgrade path to match the fresh installation.

#### `enable_notify_insert`

This function ([added here](https://github.com/pgmq/pgmq/blob/66349cfee14ab75827eb541d36b2c84cbbb68dab/pgmq-extension/sql/pgmq--1.6.1--1.7.0.sql#L9)) originally had a single parameter but was [updated to have two parameters](https://github.com/pgmq/pgmq/blob/66349cfee14ab75827eb541d36b2c84cbbb68dab/pgmq-extension/sql/pgmq--1.7.1--1.8.0.sql#L45) -- the single-parameter version does not exist in fresh installs. I'm guessing that the upgrade script was intending to replace the single-parameter version with the two-parameter version, but `CREATE OR REPLACE FUNCTION` does not replace an existing function if the parameter types are different, even if the function names match, so the single-parameter version was not removed. 

This PR removes the single parameter version. This should not be a breaking change as the two-parameter version has a default value for the second parameter, so it can be called in the same way as the old single-parameter version.

#### `pg_dump` output
First, there was a missing `PERFORM pg_catalog.pg_extension_config_dump('pgmq.topic_bindings', '');` call for the migration path. This PR adds the missing call.

Second, in the migration path, the `pgmq.notify_insert_throttle` and `pgmq.topic_bindings` tables are created after the blanket `GRANT`s from the fresh installation script, so they each have an extra `GRANT SELECT ON TABLE ... TO pg_monitor;` line in the `pg_dump` output. The fresh installation script was updated to move the blanket `GRANT`s to be before the creation of these tables as well.

Closes https://github.com/pgmq/pgmq/issues/521

### DIFF
--- a/.github/workflows/extension_upgrade.yml
+++ b/.github/workflows/extension_upgrade.yml
@@ -144,3 +144,106 @@ jobs:
         if: failure()
         working-directory: ./pgmq-extension
         run: cat regression.diffs
+
+  upgrade_scripts_do_not_diverge:
+    name: Confirm upgrade scripts do not diverge from fresh install (pg ${{ matrix.pg }})
+    runs-on: ubuntu-latest
+    container: pgxn/pgxn-tools
+    strategy:
+      fail-fast: false
+      matrix:
+        pg: [ 14, 15, 16, 17, 18 ]
+    env:
+      DATABASE_URL: postgresql://postgres@localhost:5432/postgres
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Get current branch
+        id: current-version
+        run: echo "CI_BRANCH=$(git name-rev --name-only HEAD)" >> $GITHUB_OUTPUT
+
+      - name: Start PostgreSQL ${{ matrix.pg }}
+        run: pg-start ${{ matrix.pg }}
+
+      - name: Checkout old version (v1.5.0)
+        run: git checkout tags/v1.5.0
+
+      - name: Install pgmq v1.5.0 as an extension
+        working-directory: ./pgmq-extension
+        run: |
+          make clean || true
+          make
+          make install
+          psql -c "CREATE EXTENSION pgmq;"
+
+      - name: Checkout CI branch
+        env:
+          CI_BRANCH: ${{ steps.current-version.outputs.CI_BRANCH }}
+        run: git checkout $CI_BRANCH
+
+      - name: Install and upgrade pgmq extension to CI branch version
+        working-directory: ./pgmq-extension
+        run: |
+          make clean || true
+          make
+          make install
+          psql -c "ALTER EXTENSION pgmq UPDATE;"
+
+      - name: Dump the extension schema (upgrade path)
+        run: |
+          file_name=/tmp/pg_dump_upgrade.txt
+          echo "--- pgmq extension version ---" >> $file_name
+          psql "$DATABASE_URL" -A -t -c "SELECT extversion FROM pg_extension where extname = 'pgmq';" >> $file_name
+          echo "--- pg_dump ---" >> $file_name
+          pg_dump -n pgmq -e pgmq "$DATABASE_URL" >> $file_name
+          # `pg_dump` does not dump all data, so we need to manually dump the other data we care about
+          echo "--- pgmq functions ---" >> $file_name
+          psql "$DATABASE_URL" -A -t -c "SELECT pg_get_functiondef(f.oid) FROM pg_catalog.pg_proc f INNER JOIN pg_catalog.pg_namespace n ON (f.pronamespace = n.oid) WHERE n.nspname = 'pgmq' ORDER BY pg_get_functiondef(f.oid);" >> $file_name
+          echo "--- pgmq tables ---" >> $file_name
+          psql "$DATABASE_URL" -A -c "SELECT table_name, column_name, ordinal_position, column_default, is_nullable, data_type FROM information_schema.columns WHERE table_schema = 'pgmq' ORDER BY table_name, ordinal_position;" >> $file_name
+          echo "--- pgmq types ---" >> $file_name
+          psql "$DATABASE_URL" -A -c "SELECT udt_name, attribute_name, ordinal_position, attribute_default, is_nullable, data_type FROM information_schema.attributes where udt_schema = 'pgmq' ORDER BY udt_name, ordinal_position;" >> $file_name
+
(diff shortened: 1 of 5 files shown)


ChuckHend:
This is a great addition and long overdue!

.github/workflows/extension_upgrade.yml
```diff
@@ -143,4 +143,92 @@ jobs:
...
+      - name: Drop the extension and perform a fresh installation
+        run: |
+          psql "$DATABASE_URL" -c "DROP EXTENSION pgmq;"
+          psql "$DATABASE_URL" -c "CREATE EXTENSION pgmq;"
+
+      - name: Dump the extension schema (fresh install)
+        run: |
+          echo "--- pg_dump ---" >> /tmp/pg_dump_fresh.txt
+          pg_dump -n pgmq -e pgmq "$DATABASE_URL" >> /tmp/pg_dump_fresh.txt
+          echo "--- pgmq functions ---" >> /tmp/pg_dump_fresh.txt
```
spencewenski: Todo: dump other things in the schema as well: tables, types, etc

spencewenski:
# enable_notify_insert

```sql
CREATE OR REPLACE FUNCTION pgmq.enable_notify_insert(queue_name text)
```

This function ([added here](https://github.com/pgmq/pgmq/blob/66349cfee14ab75827eb541d36b2c84cbbb68dab/pgmq-extension/sql/pgmq--1.6.1--1.7.0.sql#L9)) appears to have been [updated to have more parameters](https://github.com/pgmq/pgmq/blob/66349cfee14ab75827eb541d36b2c84cbbb68dab/pgmq-extension/sql/pgmq--1.7.1--1.8.0.sql#L45) -- the single-parameter version does not exist in fresh installs.


@ChuckHend , should we remove the single-parameter version? The two-parameter version has a default value for the second parameter, so I don't think it would be a breaking change to remove the single-parameter version.

# read_grouped_head

```sql
CREATE FUNCTION pgmq.read_grouped_head(…)
```

This function only exists in the fresh install, so it should be added via a migration file as well.

# _ensure_pg_partman_installed
This method exists for both fresh installs and upgrades, but the implementations are different. The implementation should be updated to match the fresh installation via a migration.

# pg_dump
There's also a couple differences in the `pg_dump` output that I haven't tracked down yet:

Upgrade path:
```
--
-- Name: TABLE notify_insert_throttle; Type: ACL; Schema: pgmq; Owner: postgres
--

GRANT SELECT ON TABLE pgmq.notify_insert_throttle TO pg_monitor;


--
-- Name: TABLE topic_bindings; Type: ACL; Schema: pgmq; Owner: postgres
--

GRANT SELECT ON TABLE pgmq.topic_bindings TO pg_monitor;
```

And the same lines from the fresh install:
```
--
-- Data for Name: topic_bindings; Type: TABLE DATA; Schema: pgmq; Owner: postgres
--

COPY pgmq.topic_bindings (pattern, queue_name, bound_at) FROM stdin;
\.
```

ChuckHend:
`enable_notify_insert`: we wouldn't retroactively change the v1.6.1--1.7.0 migration file for enable_notify_insert. At the time of v1.7.0, the enable_notify_insert function only had one parameter. The proper comparison to do there would be to checkout the tag v1.7.0 and validate the migration sql files up to v1.7.0 create the same state as the pgmq.sql from the tag v1.7.0.

`read_grouped_head` and `_ensure_pg_partman_installed`: agreed

`pg_dump`: postgres has different requirements for pg_dump depending on whether the objects are part of an "extension" or just regular SQL objects. What we had been going for is to handle all the extension related requirements in a block in [pgmq.sql](https://github.com/pgmq/pgmq/blob/66349cfee14ab75827eb541d36b2c84cbbb68dab/pgmq-extension/sql/pgmq.sql#L65-L75). Its possible some of these get missed in the migration SQL, since it suffers from the same issues we see elsewhere. When this is a formal "extension", Postgres only runs the code in pgmq.sql during `CREATE EXTENSION`, and the migration SQL files are only run when executing `ALTER EXTENSION pgmq UPDATE...`. Hopefully that helps clarify some things about how we see a different state from fresh install to the upgrade scripts.

spencewenski:
`enable_notify_insert`: Sorry, I probably wasn't clear -- I wasn't saying we should retroactively change the v1.6.1--1.7.0 migration file. Instead, I think we should do `DROP FUNCTION IF EXISTS pgmq.enable_notify_insert(queue_name text);` in a new migration file to remove the single-parameter version if it still exists. This should not be a breaking change as the two-parameter version has a default value for the second parameter, so it can be called in the same way as the old single-parameter version.

`pg_dump`: Yes, that makes sense. Also, I realized that I didn't update the extension version in `pgmq.control` when I did the comparison above, so the `pg_catalog.pg_extension_config_dump` calls I added in the new `pgmq--1.11.0--1.11.1.sql` migration file were not being run. I updated the version and there's still a difference, but it's a smaller difference and is closer to what I was expecting.

ChuckHend:
> DROP FUNCTION IF EXISTS pgmq.enable_notify_insert(queue_name text);

Ok, I reread your post and its more clear to me now too :)  Totally agree. I'll add a bit more detail below.

Problem is that the line below is a `CREATE OR REPLACE` but we wanted it to be a replace only.

https://github.com/pgmq/pgmq/blob/66349cfee14ab75827eb541d36b2c84cbbb68dab/pgmq-extension/sql/pgmq--1.7.1--1.8.0.sql#L45 

When this is run, its not replacing the existing single-param variant of the function. Instead, it is overloading it with the new 2 parameter variant which results in both variants existing. We want only the 2 parameter variant. So yes, we should DROP the single param variant.

ChuckHend (approved):

PR #532, opened 2026-04-20
fix: Update pgmq.send examples to avoid "function ... is not unique" error

brianpursley:
Disambiguates to avoid:
```
[42725] ERROR: function pgmq.send(unknown, unknown, unknown) is not unique
Hint: Could not choose a best candidate function. You might need to add explicit type casts.
Position: 8
```


ChuckHend (approved):

ChuckHend:
Thank you Brian.

PR #542, opened 2026-05-19
feat: Add queue insert notification support to `PGMQueueExt`

spencewenski:
Adds support for the following extension methods to `PGMQueueExt`:

- `pgmq.enable_notify_insert`
- `pgmq.disable_notify_insert`
- `pgmq.update_notify_insert`
- `pgmq.list_notify_insert_throttles`

In addition, this PR adds `queue_insert_listener*` convenience methods to create `sqlx::postgres::PgListener` instances that are registered to listen to the specified queue(s).

I also realized that sql functions that return `void` can simply be invoked using `SELECT ...` instead of `SELECT * FROM ...`, so I also aligned all relevant sql function calls to simply use `SELECT`.

### DIFF
--- a/pgmq-rs/src/pg_ext/mod.rs
+++ b/pgmq-rs/src/pg_ext/mod.rs
@@ -1,7 +1,9 @@
 mod visibility_timeout_offest;
 
 use crate::errors::PgmqError;
-use crate::types::{ListTopicBindingsRow, Message, SendBatchTopicRow, QUEUE_PREFIX};
+use crate::types::{
+    ListNotifyInsertThrottlesRow, ListTopicBindingsRow, Message, SendBatchTopicRow, QUEUE_PREFIX,
+};
 use crate::util::{check_input, connect};
 use log::info;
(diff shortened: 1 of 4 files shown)


pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -275,16 +277,16 @@ impl PGMQueueExt {
         let queue_table = format!("pgmq.{QUEUE_PREFIX}_{queue_name}");
         // we need to check whether the queue exists first
         // pg_partman create operations are currently unable to be idempotent
-        let exists_stmt = "SELECT EXISTS(SELECT * from part_config where parent_table = $1);";
-        let exists = sqlx::query_scalar(exists_stmt)
```
spencewenski, line 279: When the sqlx method call contains a static string, RustRover shows syntax highlighting for the SQL, but when the SQL is just in a string variable, no syntax highlighting is displayed.

## Before
Image

## After
Image

ChuckHend (approved):

PR #545, opened 2026-05-22
 Multi-driver support via extension trait

Sagebati:
pgmq-rs was hard-wired to sqlx. To use it with tokio-postgres, diesel-async, or diesel (sync), every consumer had to pull in sqlx
  anyway. We need more clients.

  What changed

  - PGMQueueExt is now a trait, not a struct. Implemented per-driver on the connection / transaction types directly. Users bring
  their own pool (sqlx, deadpool, bb8, mobc, custom, none).
  - Adapters added: sqlx (default), tokio-postgres, diesel-async, diesel-sync. Each behind its own cargo feature.
  - Removed _with_cxn method variants. They were sqlx-specific and didn't translate cleanly to the other drivers. Replacement:
  tx.send(...) directly, since the trait is implemented on transaction types too.
  - Removed PGMQueue struct (was already deprecated) and util::connect.
  - SQL extracted to src/sql/*.sql so all drivers share the same queries.

  Usage
```rust
  use pgmq::PGMQueueExt;

  // sqlx (default)
  let mut conn = sqlx_pool.acquire().await?;
  conn.create("orders").await?;
  conn.send("orders", &order).await?;

  // tokio-postgres
  let client = tpg_pool.get().await?;
  client.send("orders", &order).await?;

  // diesel-async
  let mut conn = diesel_pool.get().await?;
  conn.send("orders", &order).await?;

  // sync diesel
  let mut conn = PgConnection::establish(url)?;
  block_on(conn.send("orders", &order))?;

  // In your own transaction (any driver)
  let mut tx = pool.begin().await?;
  sqlx::query("INSERT INTO orders ...").execute(&mut *tx).await?;
  tx.send("orders", &order).await?;
  tx.commit().await?;
 ```

  Breaking changes (0.33.7 → 0.34.0)

  - PGMQueueExt::new(url, n) → sqlx::PgPool::connect(url) + use pgmq::PGMQueueExt;
  - _with_cxn methods removed — use the trait on &mut tx instead
  - PgmqError::DatabaseError payload: sqlx::Error → String
  - Removed PGMQueue, util::connect, PgmqError::UrlParsingError

  Bonus (separable)

  Optional tracing cargo feature adds #[tracing::instrument(skip_all)] to every queue method. Off by default. Can land independently.

### CODE REVIEW

**Comment by spencewenski** [comment]
Thanks for the contribution, @Sagebati. This is a massive PR, so it will take some time to review.

File: pgmq-rs/tests/test_sql.rs [hunk 1]
```diff
@@ -35,22 +39,27 @@ async fn test_sql_lifecycle() {
         .await
         .unwrap();
 
-    let queue = pgmq::PGMQueueExt::new(test_db_url, 1).await.unwrap();
-    #[cfg(feature = "install-sql-embedded")]
-    queue.install_sql_from_embedded().await.unwrap();
-    queue.create(&test_queue).await.unwrap();
+    let pool = PgPool::connect(&test_db_url).await.unwrap();
+    pgmq::install::sqlx::install_sql_from_embedded(&pool)
```
spencewenski, line 41: We should test installation for all drivers

pgmq-rs/tests/pg_ext_integration_test.rs
spencewenski, line 1: We need to keep all of the tests in this file. Where these tests moved somewhere? I don't see them anywhere. Presumably it should be fairly easy to refactor these tests to be abstract over the driver type.

pgmq-rs/tests/diesel_async_smoke.rs
```diff
@@ -0,0 +1,151 @@
...
+    let id: i64 = conn
+        .transaction::<_, pgmq::PgmqError, _>(|conn| {
+            async move {
+                diesel::sql_query(
+                    "CREATE TEMP TABLE IF NOT EXISTS _diesel_smoke_marker (note TEXT);",
+                )
+                .execute(conn)
+                .await?;
+                diesel::sql_query("INSERT INTO _diesel_smoke_marker (note) VALUES ('hello');")
+                    .execute(conn)
```
spencewenski: What's the point of these?

pgmq-rs/examples/basic.rs
```diff
@@ -38,54 +44,40 @@ async fn main() -> Result<(), PgmqError> {
...
+        .send(my_queue, &struct_message)
         .await
         .expect("Failed to enqueue message");
 
-    // Use a visibility timeout of 30 seconds.
-    //
-    // Messages that are not deleted within the
-    // visibility timeout will return to the queue.
-    let visibility_timeout_seconds: i32 = 30;
+    let vt = VisibilityTimeoutOffset::seconds(30);
```
spencewenski: Minor: any type that implements `Into<VisibilityTimeoutOffset>` should be able to be used as the `vt` parameter.

pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -1,1532 +1,311 @@
...
-
-    pub async fn send<T: Serialize>(
-        &self,
-        queue_name: &str,
-        message: &T,
-    ) -> Result<i64, PgmqError> {
-        self.send_with_cxn(queue_name, message, &self.connection)
-            .await
-    }
+        vt: VisibilityTimeoutOffset,
```
spencewenski: Please keep this as `impl Into<VisibilityTimeoutOffset>,`

pgmq-rs/Cargo.toml
```diff
@@ -17,60 +17,83 @@ path = "src/bin/pgmq-cli.rs"
...
-thiserror = "1.0.38"
-tokio = { version = "1.38.0", default-features = false, features = ["macros", "time"] }
-log = "0.4.18"
-url = "2.4.0"
+# 1.x+ crates pinned to just the major.
+async-trait = "0.1"
+chrono = { version = "0.4", features = ["serde"] }
+serde = { version = "1", features = ["derive"] }
+serde_json = { version = "1", features = ["raw_value"] }
+thiserror = "1"
```
spencewenski: Please restore the minor/patch versions.

pgmq-rs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.33.7"
+version = "0.34.0"
```
spencewenski: Can you change this to `0.34.0-alpha`? I think there are some other semver breaking changes I wanted to merge before releasing `0.34` (breaking changes are allowed in pre-release versions).

Or, if you’re able to change the code to not require a major version impact, that would be ideal.

pgmq-rs/src/adapters/sqlx.rs
```diff
@@ -0,0 +1,586 @@
...
+// are mechanically the same except for what `e!()` expands to (the sqlx executor used to run
+// each query). We define the body once via a macro and instantiate it twice.
+// ---------------------------------------------------------------------------------------------
+
+macro_rules! impl_pgmq_for_sqlx {
+    (
+        $target:ty,
+        |$self:ident| { $($body:tt)* }
+    ) => {
+        #[async_trait]
```
spencewenski: Is it possible to do this without a macro? Something like:

```rust
impl<T> PGMQueueExt for T where T: sqlx::Executor + sqlx::Acquire { ... }
```

I'm not sure the exact trait bounds on `T` that would be required, and I'm also not sure if it would be allowed due to rust's coherence/orphan rules for implementing traits, but this would be preferable if possible.

pgmq-rs/src/adapters/sqlx.rs
```diff
@@ -0,0 +1,586 @@
...
+                $($body)*
+                Ok(e!(sqlx::query_as::<_, QueueMetrics>(query::METRICS_ALL).fetch_all))
+            }
+        }
+    };
+}
+
+// &mut PgConnection: sqlx Executor is implemented on &mut PgConnection. Reborrow each call
+// so methods that use `e!()` multiple times still compile.
+impl_pgmq_for_sqlx!(&mut PgConnection, |self| {
```
spencewenski: It would be great if we could implement for a pool as well. Same comment for the other adapters.

pgmq-rs/src/adapters/diesel_async.rs
```diff
@@ -0,0 +1,820 @@
...
+    }
+}
+
+// ---------------------------------------------------------------------------------------------
+// Scalar return-column structs — one per shape that diesel needs to decode.
+// ---------------------------------------------------------------------------------------------
+
+#[derive(QueryableByName)]
+struct SendCol {
+    #[diesel(sql_type = sql_types::BigInt)]
```
spencewenski: Hmm, I'm pretty sure the `#[diesel(sql_type = ...)]` annotations are not needed, here and elsewhere throughout this PR.

pgmq-rs/src/adapters/diesel_async.rs
```diff
@@ -0,0 +1,820 @@
...
+mod imp {
+    use super::*;
+
+    #[cfg_attr(feature = "tracing", tracing::instrument(skip_all))]
+    pub(super) async fn create(
+        conn: &mut AsyncPgConnection,
+        queue_name: &str,
+    ) -> Result<(), PgmqError> {
+        check_input(queue_name)?;
+        sql_query(query::CREATE)
```
spencewenski: If possible, it would be great to use the `sql_function!()` macro to define rust functions for diesel to use. These could also be exposed as part of `pgmq` public crate API.

pgmq-rs/src/sql/delete.sql
```diff
@@ -0,0 +1 @@
+SELECT * from pgmq.delete(queue_name=>$1::text, msg_id=>$2::bigint) AS delete;
```
spencewenski: Is it a problem that `delete` is a SQL keyword?

pgmq-rs/src/errors.rs
```diff
@@ -9,13 +8,17 @@ pub enum PgmqError {
     #[error("json parsing error {0}")]
     JsonParsingError(#[from] serde_json::error::Error),
 
-    /// a url parsing error
-    #[error("url parsing error {0}")]
-    UrlParsingError(#[from] ParseError),
+    /// a database error returned by the underlying driver
+    #[error("database error: {0}")]
+    DatabaseError(String),
```
spencewenski: It's unfortunate to lose the reference to the actual DB error. Could we instead have a different variant for each possible error type?

pgmq-rs/src/adapters/helpers.rs
```diff
@@ -0,0 +1,50 @@
...
+//!
+//! Items are `pub` so sibling adapter modules can use them via `super::helpers::*`.
+
+use crate::errors::PgmqError;
+use serde::Serialize;
+
+/// Default poll timeout for `read_with_poll` family methods, in seconds.
+pub const DEFAULT_POLL_TIMEOUT_S: i32 = 5;
+/// Default poll interval for `read_with_poll` family methods, in milliseconds.
+pub const DEFAULT_POLL_INTERVAL_MS: i32 = 250;
```
spencewenski: IIRC the extension methods all provide default values if these aren't provided. It would be great if we could use the extension's defaults instead of defining (potentially different) defaults in the client.

pgmq-rs/src/adapters/diesel_async.rs
```diff
@@ -0,0 +1,820 @@
...
+// Both impls (Pool, &mut AsyncPgConnection) delegate every method to `imp::` shared body,
+// varying only in how they obtain the `&mut AsyncPgConnection` to pass in. Macro factors out the
+// duplication. Methods take `self` by value; auto-(re)borrow handles the reference passing.
+// ---------------------------------------------------------------------------------------------
+
+macro_rules! impl_pgmq_for_diesel {
+    ($target:ty, |$self:ident| $conn:expr) => {
+        #[async_trait]
+        impl PGMQueueExt for $target {
+            async fn create($self, queue_name: &str) -> Result<(), PgmqError> { imp::create($conn, queue_name).await }
```
spencewenski: Can you reformat the code in this macro?

pgmq-rs/src/adapters/diesel_async.rs
```diff
@@ -0,0 +1,820 @@
...
+            async fn send_batch_with_delay_with_headers<T: Serialize + Send + Sync, H: Serialize + Send + Sync>($self, queue_name: &str, messages: &[T], headers: Option<&[H]>, delay: VisibilityTimeoutOffset) -> Result<Vec<i64>, PgmqError> { imp::send_batch_with_delay_with_headers($conn, queue_name, messages, headers, delay).await }
+            async fn read<T: for<'de> Deserialize<'de> + Send + Unpin + 'static>($self, queue_name: &str, vt: VisibilityTimeoutOffset) -> Result<Option<Message<T>>, PgmqError> { Ok($self.read_batch::<T>(queue_name, vt, 1).await?.into_iter().next()) }
+            async fn read_batch<T: for<'de> Deserialize<'de> + Send + Unpin + 'static>($self, queue_name: &str, vt: VisibilityTimeoutOffset, qty: i32) -> Result<Vec<Message<T>>, PgmqError> { imp::read_batch::<T>($conn, queue_name, vt, qty).await }
+            async fn read_with_poll<T: for<'de> Deserialize<'de> + Send + Unpin + 'static>($self, queue_name: &str, vt: VisibilityTimeoutOffset, pt: Option<std::time::Duration>, pi: Option<std::time::Duration>) -> Result<Option<Message<T>>, PgmqError> { Ok($self.read_batch_with_poll::<T>(queue_name, vt, 1, pt, pi).await?.and_then(|v| v.into_iter().next())) }
```
spencewenski: Please use full variable names instead of abbreviations (`poll_timeout` instead of `pt`). The exception being `vt` -- however, I wouldn't be opposed to changing that to `visibility_timeout` as well, here and in the trait definition.

pgmq-rs/Cargo.toml
```diff
@@ -17,60 +17,83 @@ path = "src/bin/pgmq-cli.rs"
...
-# Optional dependencies for the `cli` feature
-clap = { version = "4.1", features = ["derive"], optional = true }
-env_logger = { version = "0.10.0", optional = true }
+# Optional driver-adapter dependencies.
+sqlx = { version = "0.8", default-features = false, features = ["runtime-tokio", "postgres", "chrono", "json", "macros"], optional = true }
+tokio-postgres = { version = "0.7", default-features = false, features = ["with-chrono-0_4", "with-serde_json-1"], optional = true }
+# deadpool-postgres is a *dev-dep only* — used by smoke tests/examples. Users bring whatever
+# pool they prefer (deadpool, bb8, mobc, custom, none). pgmq depends only on tokio-postgres.
+diesel = { version = "2", default-features = false, features = ["postgres", "chrono", "serde_json"], optional = true }
+diesel-async = { version = "0.5", default-features = false, features = ["postgres", "deadpool"], optional = true }
```
spencewenski: `0.5` is very old. Please upgrade to `0.9`

pgmq-rs/Cargo.toml
```diff
@@ -17,60 +17,83 @@ path = "src/bin/pgmq-cli.rs"
...
+thiserror = "1"
 
-# Optional dependencies for the `cli` feature
-clap = { version = "4.1", features = ["derive"], optional = true }
-env_logger = { version = "0.10.0", optional = true }
+# Optional driver-adapter dependencies.
+sqlx = { version = "0.8", default-features = false, features = ["runtime-tokio", "postgres", "chrono", "json", "macros"], optional = true }
+tokio-postgres = { version = "0.7", default-features = false, features = ["with-chrono-0_4", "with-serde_json-1"], optional = true }
+# deadpool-postgres is a *dev-dep only* — used by smoke tests/examples. Users bring whatever
+# pool they prefer (deadpool, bb8, mobc, custom, none). pgmq depends only on tokio-postgres.
```
spencewenski: These comment seems out of place

pgmq-rs/Cargo.toml
```diff
@@ -17,60 +17,83 @@ path = "src/bin/pgmq-cli.rs"
...
+diesel = { version = "2", default-features = false, features = ["postgres", "chrono", "serde_json"], optional = true }
```
spencewenski: Please provide specific minor/patch versions

pgmq-rs/Cargo.toml
```diff
@@ -17,60 +17,83 @@ path = "src/bin/pgmq-cli.rs"
...
-
-# `pgmq` only uses the `chrono` feature of `sqlx`, so it's possible to add a query that would encounter this conflict
-# in a consumer and not notice because `pgmq` compiles fine on its own. To ensure queries that would see this conflict
-# are not added in the future, add the `time` feature in dev dependencies. This is a "regression test" of sorts, in
-# that it will enable the `time` feature in tests/examples, simulating the behavior of a consumer of `pgmq` that has
-# both the `time` and `chrono` features enabled. If a query that would see the conflict is added, `pgmq` will fail to
-# compile when running tests/examples.
-sqlx = { version = "0.8.1", features = ["time"] }
+# Note: default-features = false to avoid pulling in sqlx::Any, which causes Type<_> ambiguity errors.
+sqlx = { version = "0.8", default-features = false, features = ["time", "postgres", "runtime-tokio", "chrono", "macros", "json"] }
```
spencewenski: Why is this change necessary?

pgmq-rs/src/adapters/query.rs
```diff
@@ -0,0 +1,70 @@
...
+/// this as a static string. Returns the final SQL plus an ordered list of optional params to bind.
+pub fn convert_archive_partitioned_sql(
+    has_partition_interval: bool,
+    has_retention_interval: bool,
+) -> String {
+    use std::fmt::Write;
+    let mut sql = String::from("SELECT pgmq.convert_archive_partitioned(table_name=>$1::text");
+    let mut idx = 2;
+    if has_partition_interval {
+        // `write!` to a String never fails — unwrap is safe.
```
spencewenski: Hmm, is this documented somewhere? Blanket statements like this make me nervous if they aren't backed by documentation.

pgmq-rs/src/adapters/sqlx.rs
```diff
@@ -0,0 +1,586 @@
...
+    macro_rules! e { ($($t:tt)*) => { $($t)*(&mut *self).await? } }
+});
+
+// &mut Transaction<'_, Postgres>: sqlx Executor accepts it via DerefMut to PgConnection.
+impl_pgmq_for_sqlx!(&mut Transaction<'_, Postgres>, |self| {
+    macro_rules! e { ($($t:tt)*) => { $($t)*(&mut **self).await? } }
```
spencewenski: Can we use a more descriptive name than `e`?

pgmq-rs/src/install/diesel_async.rs
spencewenski, line 1: There appears to be a lot of duplication of the installation logic between the various drivers, which will make it difficult to maintain. Can we consolidate the logic with some abstractions?

pgmq-rs/tests/diesel_async_smoke.rs
```diff
@@ -0,0 +1,151 @@
...
+
+async fn pool() -> Pool<AsyncPgConnection> {
+    let pool = pool_from_env();
+    pgmq::install::diesel_async::install_sql_from_embedded(&pool)
+        .await
+        .expect("install pgmq");
+    pool
+}
+
+#[ignore]
```
spencewenski: I'm not sure I agree that these tests should be marked with `#[ignore]`.

pgmq-rs/tests/diesel_async_smoke.rs
```diff
@@ -0,0 +1,151 @@
+//! End-to-end smoke test for the diesel-async adapter.
+//!
+//! Run with:
+//!   DATABASE_URL=postgres://postgres:postgres@localhost:5432/postgres \
+//!   cargo test --no-default-features --features diesel-async,install-sql-embedded \
+//!     --test diesel_async_smoke -- --ignored --nocapture
+
+#![cfg(all(feature = "diesel-async", feature = "install-sql-embedded"))]
```
spencewenski: We need to test with a normal extension installation as well.

pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -1,1532 +1,311 @@
...
-    ) -> Result<sqlx::Transaction<'c, Postgres>, PgmqError> {
-        let txn = self
-            .acquire_queue_lock_with_cxn(queue_name, &self.connection)
-            .await?;
-        Ok(txn)
-    }
+/// Queue API for the `pgmq` Postgres extension. Implemented natively by each driver adapter.
+/// Bring this trait into scope to call queue methods directly on your pool or transaction.
+#[async_trait]
+pub trait PGMQueueExt {
```
spencewenski: I'd suggest keeping the existing `PGMQueueExt` struct (you can mark it deprecated) and using a different name for the trait (maybe something simple like `Queue`). We've been discussing internally that we would prefer to use a different name than `PGMQueueExt` for this struct, so making the name change with a new trait makes sense. This would also allow this change to be non-api breaking.

pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -1,1532 +1,311 @@
...
 
-    pub async fn list_notify_insert_throttles(
-        &self,
-    ) -> Result<Vec<ListNotifyInsertThrottlesRow>, PgmqError> {
-        self.list_notify_insert_throttles_with_cxn(&self.connection)
-            .await
-    }
+/// sqlx-specific notification listener helpers.
+#[cfg(feature = "sqlx")]
+pub mod sqlx_listener {
```
spencewenski: This should probably go in a sqlx-specific mod. Additionally, if the other drivers have a "listener" concept, it would be nice to put this functionality in a trait as well. It could be a separate trait, maybe `QueueListener` or something. At minimum, we should provide similar convenience methods for each driver where it's possible.

pgmq-rs/src/sql/install/init.sql
```diff
@@ -0,0 +1,8 @@
+-- Used by install::*::init(). Combines the advisory lock + CREATE EXTENSION into one
+-- multi-statement batch (single roundtrip via the simple-query protocol).
+--
+-- The advisory lock key (-9223372036854771659 = i64::MIN + 4149) is a randomly-chosen large
+-- negative bigint, picked to minimize collision with application-level advisory locks. See
+-- src/install/mod.rs::ADVISORY_LOCK_KEY.
+SELECT pg_advisory_xact_lock(-9223372036854771659);
```
spencewenski: I'd prefer to keep a single constant in code for the advisory lock value.

pgmq-rs/src/pg_ext/visibility_timeout_offest.rs
```diff
@@ -152,18 +145,3 @@ impl From<std::time::Duration> for VisibilityTimeoutOffset {
...
-*/
-impl<DB: Database> Type<DB> for VisibilityTimeoutOffset
-where
-    i32: Type<DB>,
-    std::time::Duration: Type<DB>,
-{
-    fn type_info() -> DB::TypeInfo {
-        <i32 as sqlx::Type<DB>>::type_info()
-    }
-}
```
spencewenski, line 169: Why was this removed?

pgmq-rs/src/pg_ext/visibility_timeout_offest.rs
```diff
@@ -81,7 +74,7 @@ use std::ops::Deref;
 /// # use pgmq::pg_ext::VisibilityTimeoutOffset;
 /// assert_eq!(VisibilityTimeoutOffset::MAX, VisibilityTimeoutOffset::from(std::time::Duration::MAX));
 /// ```
-#[derive(Clone, Copy, Debug, PartialEq, Encode)]
+#[derive(Clone, Copy, Debug, PartialEq)]
```
spencewenski, line 86: Why was `Encode` removed?

pgmq-rs/src/pg_ext/visibility_timeout_offest.rs
```diff
@@ -64,12 +63,6 @@ use std::ops::Deref;
 /// assert_eq!(VisibilityTimeoutOffset::MAX, VisibilityTimeoutOffset::from(chrono::Duration::MAX));
 /// ```
 ///
-/// ## Convert from [`chrono::Duration`] -- capped min
-/// ```
-/// # use pgmq::pg_ext::VisibilityTimeoutOffset;
-/// assert_eq!(VisibilityTimeoutOffset::MAX, VisibilityTimeoutOffset::from(chrono::Duration::MAX));
-/// ```
-///
```
spencewenski, line 72: Why was this example removed?

pgmq-rs/src/pg_ext/visibility_timeout_offest.rs
```diff
@@ -1,9 +1,8 @@
-use sqlx::{Database, Encode, Type};
 use std::ops::Deref;
 
 /// Type to represent an offset that will be applied to the current timestamp in order to update
 /// the `vt` (visibility timeout) timestamp column of a queue message. Used by various
-/// methods in [`crate::pg_ext::PGMQueueExt`] to set the visibility timeout of a job. Supports
+/// methods in [`crate::PGMQueueExt`] to set the visibility timeout of a job. Supports
```
spencewenski: Hmm, this import doesn't look correct. I don't see a `pub use` in `lib.rs` that would allow this import to work. Unless I'm missing something?

spencewenski:
@Sagebati  I left an initial round of comments. At a high level I think I'm aligned with the overall direction of this PR, however, it seems to have been largely generated by AI as there's a lot of slop and weird changes that don't seem like they were necessary or intentional. Please address the initial round of comments, thoroughly review the PR yourself and fix anything that seems weird/unnecessary to you, and let me know when you're ready for a full review.

pgmq-rs/Cargo.toml
Sagebati: I do understand this is not related to the PR  and i'll execute myself.

but why do we need the minor/patch versions ? mentioning patch version it pure aesthetic. But not specifying minor could improve the resolver.

pgmq-rs/src/adapters/sqlx.rs
Sagebati: The first design the goal was to implement it on the pools directly. But the implementation become repetitive and cumber-stone. The DX gains are not worth it, in my opinion.

pgmq-rs/src/adapters/diesel_async.rs
Sagebati: Unfortunately they are. https://docs.rs/diesel/latest/diesel/deserialize/derive.QueryableByName.html

They would not be if I used the schema!{} macro from diesel. Maybe for a next PR.

pgmq-rs/src/adapters/diesel_async.rs
Sagebati: I think it's a good idea, I assumed a MVP version without all the machinery of diesel was easier to ship. I think adding it later will be forward compatible

pgmq-rs/src/errors.rs
Sagebati: Yes!

pgmq-rs/src/adapters/diesel_async.rs
Sagebati: Ok

pgmq-rs/src/install/diesel_async.rs
Sagebati: I really tried to have a common abstraction, one of the limitation is the way they execute SQL (&mut ref, &ref) and the transactions. Maybe I can give it another try

pgmq-rs/src/pg_ext/visibility_timeout_offest.rs
Sagebati: I'll revert

Sagebati:
Hi,

Thanks for the review.

I've reworked the PR with the feedback in mind. I think it's ready for another pass.

A few questions / notes:

- I reverted the patch-version removal in Cargo.toml. I'd like to understand why pinning the patch is needed. Keeping the versions flexible should be fine with resolver ≥ 3.
- I couldn't find any cleaner cross-adapter abstraction. We could imagine a trait that abstracts over any DB connection–like type, but I think that's a dead end.
- The diesel-based implementation idea is a good one, but I think it's out of scope here. Possibly something for a different crate.
- I also removed the AI-generated smoke tests. Happy to re-add them if you'd like.

pgmq-rs/src/adapters/sqlx.rs
```diff
@@ -0,0 +1,683 @@
...
+
+/// Sealed marker trait — restricts the blanket [`Queue`] impl below to the three sqlx
+/// types we want to support, without conflicting with the per-driver impls in
+/// `tokio_postgres` / `diesel_async` / `diesel_sync` adapters.
+mod sealed {
+    pub trait SqlxConn {}
+    impl SqlxConn for &sqlx::PgPool {}
+    impl SqlxConn for &mut sqlx::PgConnection {}
+    impl SqlxConn for &mut sqlx::Transaction<'_, sqlx::Postgres> {}
+}
```
spencewenski, line 38: I don't think this does that it's claiming. The only reason this isn't giving an error currently is because the other adapters are only implemented for a single concrete type instead of as a blanket impl for any diesel, etc connection type.

spencewenski:
@Sagebati , honestly, there's no way we can reasonably review >6000 lines all at once, especially because they seem to be mostly or entirely generated by Claude, and I can tell from a brief scan that there's a lot of weird/inconsistent stuff, or stuff that doesn't quite work the way a comment says it does, so we would need to carefully review each line.

If you want to continue working on this, it will need to be split up into smaller, more manageable PRs. Perhaps, start with one or two methods on the `Queue` trait and implement for one or two drivers so we can align on things like how we want to implement for each adapter (e.g. a blanket impl vs macros as you had previously done it) without needing to worry about reviewing/approving thousands of lines all at once. You could keep the Queue trait and its impls as `pub(crate)` until it's ready for external use so we're not blocked on releasing other semver-compatible changes in the mean time.

Also, this work was on my radar to pick up soon-ish, if you'd prefer to simply wait for it to be done by us.

pgmq-rs/Cargo.toml
spencewenski: It's not just aesthetic for library crates -- lib crates should specify the minimum version they depend on.

See:

- https://github.com/taiki-e/cargo-minimal-versions
- https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#direct-minimal-versions
- https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#minimal-versions

Sagebati:
Thanks for your time. 

I'll close it. And wait for your implementation

Sagebati:
I agree with this that's why I removed the minor version so we don't depend
on features that we don't need accidentally.

Maybe I misunderstood something

Le mer. 27 mai 2026, 09:13, Spencer Ferris ***@***.***> a
écrit :

PR #546, opened 2026-05-22
Add delete notifications

clint-latch:
Mirrors the existing notify_insert mechanism for queue delete events

### DIFF
--- a/pgmq-extension/client-tests/basic_notify_delete_test.py
+++ b/pgmq-extension/client-tests/basic_notify_delete_test.py
@@ -1,0 +1,117 @@
+import json
+import os
+import pytest
+import time
+
+import psycopg
+
+
+def listen_for_notification(
+    db_connection: psycopg.Connection, channel_name: str, timeout: int = 5
+) -> bool:
+    # wait for notification with timeout using the built-in timeout
+    try:
+        for notification in db_connection.notifies(timeout=timeout):
+            print(
+                f"Received notification: channel={notification.channel}, payload={notification.payload}"
+            )
+            assert notification.channel == channel_name, (
+                f"Expected channel {channel_name}, got {notification.channel}"
+            )
+            return True
+    except TimeoutError:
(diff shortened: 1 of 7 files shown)


spencewenski:
@clint-latch , can you elaborate on why this is needed? The insert notifications are already a pretty limited use case, and I’m struggling to think of a use case where delete notifications would be useful.

clint-latch:
> @clint-latch , can you elaborate on why this is needed? The insert notifications are already a pretty limited use case, and I’m struggling to think of a use case where delete notifications would be useful.

I have a use case for a message queue. Adding this enabled back-pressure-aware consumers: when a reader is processing as fast as it can and the queue has more messages than available worker slots, the DELETE
  notification lets the reader immediately refill the slot the moment one frees up, rather than relying on a polling interval.

While stress testing the message queue it would lock up without this, unless I polled which slowed down the throughput

spencewenski:
> > @clint-latch , can you elaborate on why this is needed? The insert notifications are already a pretty limited use case, and I’m struggling to think of a use case where delete notifications would be useful.
> 
> I have a use case for a message queue. Adding this enabled back-pressure-aware consumers: when a reader is processing as fast as it can and the queue has more messages than available worker slots, the DELETE notification lets the reader immediately refill the slot the moment one frees up, rather than relying on a polling interval.
> 
> While stress testing the message queue it would lock up without this, unless I polled which slowed down the throughput

Hmm, maybe I’m not understanding something. Couldn’t the worker fetch from the queue directly once it’s free, or send some sort of in-process notification to your worker/task manager that it’s ready for a new job?

I’m not necessarily opposed to adding this, btw, just trying to understand the use case before we sign up to support it.

clint-latch:
@spencewenski thanks for pushing back. The same logic can be achieved by getting the worker to re-read the queue. I did some more testing and it achieves the same performance as using the delete notification. I'll close this PR for now.

PR #549, opened 2026-05-30
feat!: Initial `Queue` trait and `sqlx` implementation

spencewenski:
Initial commit to add a new `Queue` trait to allow supporting multiple different SQL clients (this PR only adds the `sqlx` implementation).

Some high-level details:

- The new `Queue` trait and implementations are in the `queue` mod
- Only three key methods are added initially: `create`, `send`, and `read`
- Common implementation details exist in a macro, and individual implementations are generated by invoking the macro with the relevant parameters
- Added a `sqlx` feature flag for the `PGMQueueExt` client and the `sqlx` implementation of the `Queue` trait
- Tests for all the implementations of the `Queue` trait are in the new `queue` integration test file. Each implementation has a separate test method generated for it using a new custom test macro.

Implementations for other clients (diesel/rust-postgres) will be added in a separate followup PR.

### DIFF
--- a/pgmq-rs/src/types/mod.rs
+++ b/pgmq-rs/src/types/mod.rs
@@ -1,4 +1,8 @@
 pub mod queue_name;
+pub mod visibility_timeout_offset;
+
+pub use queue_name::QueueName;
+pub use visibility_timeout_offset::VisibilityTimeoutOffset;
 
 use serde::Deserialize;
 use sqlx::types::chrono::{DateTime, Utc};
--- a/pgmq-rs/src/types/mod.rs
+++ b/pgmq-rs/src/types/mod.rs
@@ -4,9 +4,8 @@ pub mod visibility_timeout_offset;
 pub use queue_name::QueueName;
 pub use visibility_timeout_offset::VisibilityTimeoutOffset;
 
-use serde::Deserialize;
-use sqlx::types::chrono::{DateTime, Utc};
-use sqlx::FromRow;
+use chrono::{DateTime, Utc};
+use serde_derive::Deserialize;
 use std::time::Duration;
 
 pub const VT_DEFAULT: i32 = 30;
@@ -18,7 +17,8 @@ pub const QUEUE_PREFIX: &str = r#"q"#;
 pub const ARCHIVE_PREFIX: &str = r#"a"#;
 pub const PGMQ_SCHEMA: &str = "pgmq";
 
-#[derive(Clone, Debug, Deserialize, FromRow)]
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(feature = "sqlx", derive(sqlx::FromRow))]
 #[non_exhaustive]
 pub struct PGMQueueMeta {
     pub queue_name: String,
@@ -31,7 +31,8 @@ pub struct PGMQueueMeta {
 ///
 /// It is an "envelope" for the message that is stored in the queue.
 /// It contains both the message body but also metadata about the message.
-#[derive(Clone, Debug, Deserialize, FromRow)]
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(feature = "sqlx", derive(sqlx::FromRow))]
 #[non_exhaustive]
 pub struct Message<T = serde_json::Value, H = serde_json::Value> {
     /// Unique identifier for the message.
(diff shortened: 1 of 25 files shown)


pgmq-rs/src/queue/tokio_postgres.rs
```diff
@@ -0,0 +1,99 @@
...
+    visibility_timeout: VisibilityTimeoutOffset,
+    quantity: i32,
+) -> Result<Vec<Message<T>>, PgmqError>
+where
+    C: tokio_postgres::GenericClient,
+    T: Send + for<'de> serde::Deserialize<'de>,
+{
+    let row = executor
+        .query_typed(
+        // Todo: How to deal with optional headers/delay?
```
l-7-l: sorry, seems like:

```rust
let params: Vec<&(dyn ToSql+Sync)>=vec![
                (&*queue_name, tokio_postgres::types::Type::TEXT),
                (&*visibility_timeout, tokio_postgres::types::Type::INT4),
                (&quantity, tokio_postgres::types::Type::INT4),
];
let mut temp_stmt = String::new();
if let Some(seconds) = &delay {
 params.push(seconds);
let len = params.len();
temp_stmt.push_str(",delay=>$4::int4"); // or temp_stmt.push_str(&format!(",delay=>${len}::int4"));
}
let stmt = format!("SELECT msg_id, read_ct, enqueued_at, vt, message from pgmq.read(queue_name=>$1::text, vt=>$2::integer, qty=>$3::integer${temp_stmt})" );
```

pgmq-rs/src/types/mod.rs
```diff
@@ -1,9 +1,10 @@
-use chrono::serde::ts_seconds::deserialize as from_ts;
-use serde::Deserialize;
-use sqlx::types::chrono::{DateTime, Utc};
-use sqlx::FromRow;
+use chrono::{DateTime, Utc};
```
l-7-l, line 7: Maybe we can make chrono optional. 
But we should assert that at least one time library is retained — for example, either chrono or jiff.
```rust
#[cfg(feature="with-chrono-0_4")]
```

l-7-l: I'm a bit pressed for time, so progress may be slow, but I'm happy to help.

pgmq-rs/src/types/mod.rs
```diff
@@ -23,59 +24,65 @@ pub struct PGMQueueMeta {
 ///
 /// It is an "envelope" for the message that is stored in the queue.
 /// It contains both the message body but also metadata about the message.
-#[derive(Clone, Debug, Deserialize, FromRow)]
+// Todo: Add missing fields from the SQL extension's type. This would be a semver-breaking change.
+#[derive(Clone, Debug, Deserialize)]
+#[cfg_attr(feature = "sqlx", derive(sqlx::FromRow))]
+#[cfg_attr(feature = "diesel", derive(diesel::FromSqlRow))]
```
l-7-l: Sorry, since rust-postgres supports chrono and serde, perhaps we can directly use #[cfg_attr(feature="postgres", derive(postgres_types::FromSql, postgres_types::ToSql))] here.

spencewenski: Ooh, thanks, I'll try that. I'm not too familiar with the  `postgres`/`tokio-postgres` crates, so this is helpful.

spencewenski: Done. I still need to complete the rust-postgres support and add tests, but it seems to compile with this. Note that I don't think there's currently a need for the `ToSql` implementation for this type -- we don't directly insert a full message row from this crate.

pgmq-rs/src/types/mod.rs
spencewenski: Hmm, I'm not necessarily opposed to this, but I think this would be outside the scope of this PR. I'm also not entirely sure how to support multiple time types at the moment. Feel free to open a separate issue for this if there isn't one already.

pgmq-rs/src/types/mod.rs
spencewenski: Actually, it seems that the `FromSql` trait is only used to translate a single column from SQL to Rust, not an entire row. Instead, unless I'm missing something, the columns need to be manually extracted using the `Row::try_get(...)` method. I added a `impl TryFrom<Row> for Message<T>` that does this.

ChuckHend (approved):

PR #554, opened 2026-06-23
chore!: Remove deprecated `PGMQueue` rust client struct

spencewenski:


pgmq-rs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pgmq"
-version = "0.33.7"
+version = "0.34.0-alpha.0"
```
spencewenski, line 3: Creating an "alpha" pre-release version so we can group some other breaking changes together before bumping to `0.34.0` (breaking changes are allowed for pre-release versions).

pgmq-rs/src/util.rs
```diff
@@ -35,25 +35,6 @@ pub async fn connect(url: &str, max_connections: u32) -> Result<Pool<Postgres>,
...
-    match row {
-        Ok(row) => Ok(Some(row)),
-        Err(sqlx::error::Error::RowNotFound) => Ok(None),
-        Err(e) => Err(e)?,
-    }
-}
-
 /// A string that is known to be formed of only ASCII alphanumeric or an underscore;
 #[derive(Clone, Copy)]
 pub struct CheckedName<'a>(&'a str);
```
spencewenski, line 38: `CheckedName` is also not used anymore. However, I plan to rename it in https://github.com/pgmq/pgmq/pull/549, so I'll keep it here for now.

ChuckHend (approved):

PR #556, opened 2026-06-25
feat!: Add `headers` and `last_read_at` fields to `Message` struct

### LINKED ISSUE #548: Expose headers when reading a message
In the website documentation of the SQL API, the `headers` column is returned, but not available in the `Message` Rust struct.

We're storing a `public-id` header for handling webhooks in third party services. Headers seems a good fit rather than being stored in the message.

Thanks.

spencewenski:
Also:
- Mark `Message` and all other structs with only public fields as `#[non_exhaustive]` so additional fields can be added in the future as a non-semver breaking change
- Consolidate `PGMQMetadata` structs and add `sqlx::FromRow` derive to simplify deserializing it from a `PgRow`.

Resolves: https://github.com/pgmq/pgmq/issues/548

### DIFF
--- a/pgmq-rs/src/types.rs
+++ b/pgmq-rs/src/types.rs
@@ -1,4 +1,3 @@
-use chrono::serde::ts_seconds::deserialize as from_ts;
 use serde::Deserialize;
 use sqlx::types::chrono::{DateTime, Utc};
 use sqlx::FromRow;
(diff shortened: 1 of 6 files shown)


pgmq-rs/src/types.rs
```diff
@@ -24,19 +26,24 @@ pub struct PGMQueueMeta {
 /// It is an "envelope" for the message that is stored in the queue.
 /// It contains both the message body but also metadata about the message.
 #[derive(Clone, Debug, Deserialize, FromRow)]
-pub struct Message<T = serde_json::Value> {
-    /// unique identifier for the message
+#[non_exhaustive]
+pub struct Message<T = serde_json::Value, H = serde_json::Value> {
+    /// Unique identifier for the message.
     pub msg_id: i64,
-    #[serde(deserialize_with = "from_ts")]
```
spencewenski, line 30: This `#[serde(deserialize_with = "from_ts")]` didn't seem necessary, but I can add it back if there's a use case for it I'm not thinking of

ChuckHend (approved):
Awesome!

PR #560, opened 2026-07-02
feat!: Refactor `CheckedName`/`check_input`

spencewenski:
Minor refactor of `CheckedName` struct and `check_input` function into `QueueName` struct and `check_queue_name` function.

- Place the new struct and method in the new `types/queue_name` module
- Create private sub-module to contain the `MAX_PGMQ_QUEUE_LEN` constant and the intermediary constants used to calculate its value
- Add missing calls to `check_queue_name` in `PGMQueueExt` methods
- Impl `AsRef` and `Deref` for `QueueName` to make it easy to use as a paramter in SQL commands with a simple deref (e.g., `*queue_name`)
- Define a new `QueueNameError` enum to provide more detailed error responses for invalid queue names
- Add dependency on `rstest`, a crate that provides macros to help with writing parameterized tests
- Add tests for the checks in `check_queue_name`, using `rstest` to help with testing multiple different inputs using shared test logic

### DIFF
--- a/pgmq-rs/src/types/queue_name.rs
+++ b/pgmq-rs/src/types/queue_name.rs
@@ -1,0 +1,161 @@
+use constants::MAX_PGMQ_QUEUE_LEN;
+use std::fmt::{Display, Formatter};
+use std::ops::Deref;
+use thiserror::Error;
+
+/// Private module to contain constants used to calculate [`MAX_PGMQ_QUEUE_LEN`]. Intermediary
+/// constants are kept private to the module to ensure they are not referenced by accident.
+///
+/// Relevant docs: <https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS>
+mod constants {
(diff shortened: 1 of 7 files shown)


pgmq-rs/src/types/queue_name.rs
```diff
@@ -0,0 +1,161 @@
...
+impl QueueNameError {
+    /// Construct [`QueueNameError::Other`] from the given error/message.
+    pub fn other(err: impl ToString) -> Self {
+        Self::Other(err.to_string())
+    }
+}
+
+/// A valid queue name.
+#[derive(Debug, Clone, Copy)]
+pub struct QueueName<'a>(&'a str);
```
spencewenski, line 47: `QueueName` isn't used anywhere yet, but I plan to use it in https://github.com/pgmq/pgmq/pull/549. I pulled it out into this separate PR in order to make #549 smaller.

ChuckHend (approved):

PR #563, opened 2026-07-06
feat: Add diesel implementations of the `Queue` trait

spencewenski:
This PR adds support for `diesel` and `diesel-async` implementations of the `Queue` trait. Similar to the rust-postgres implementations, the sync and async implementations for diesel are nearly identical -- the only differences are which `RunQueryDsl` trait is used and that one requires `await`-ing DB results. This means, similar to the rust-postgres implementation, we can use a custom macro to generate the required code from the same shared logic.

A note about the diesel implementation: while `diesel` provides a way to execute raw SQL queries similar to `sqlx`/rust-postgres, it also has a more advanced DSL that provides better type safety at compile time. For example, we can generate Rust versions of SQL functions using the `declare_sql_function` macro that can be used used to invoke the SQL function with types checked at compile time. Using this approach means that the diesel queries are maintained separately from the `sqlx`/rust-postgres query strings; however, in my opinion benefit of the additional type safety outweighs the downside of the duplication. An additional benefit of using the `declare_sql_function` macro is the functions can be exported from the `pgmq` crate to provide direct access to typed versions of the the PGMQ functions to consumers if they don't want to use the `Queue` trait.


ChuckHend:
>  the diesel queries are maintained separately from the sqlx/rust-postgres query strings

are those queries in this PR or are they coming in later? I was looking for the duplication but can't seem to find it.

spencewenski:
> > the diesel queries are maintained separately from the sqlx/rust-postgres query strings
> 
> are those queries in this PR or are they coming in later? I was looking for the duplication but can't seem to find it.

The queries I’m referring to are in `pgmq-rs/src/queue/diesel/query.rs` and the related functions in `pgmq-rs/src/queue/diesel/sql.rs`. They’re typed queries with diesel’s DSL vs the string queries used by sqlx and rust-postgres.

The duplication in this context is that the diesel implementation uses a different approach to construct queries vs reusing the query strings from the sqlx and rust-postgres implementations. If you would prefer to reuse the query strings, let me know; it would be possible to reuse them using diesel’s `sql_query` function.

ChuckHend (approved):
@spencewenski, ok I see now. I think that makes sense to use `declare_sql_function` in this case. No need to change it. Thanks for explaining.

PR #567, opened 2026-07-14
feat: Add `Queue#delete` method and implement for all sql drivers

spencewenski:
This PR adds the `archive` method to the `Queue` trait and adds implementations for all the sql drivers we support (sqlx, diesel, rust-postgres). Use the sqlx implementation to implement `PGMQueueExt#delete_batch`.


ChuckHend (approved):
Awesome

Issue #127, opened 2023-09-27
Streamline pgmq-rs ext client queue creation behaviour


v0idpwn:

v0idpwn:
The extension-less client returns `Result<(), Error>`, which is simpler and probably as useful, since the inforrmation about the queue being new or not is not that relevant.

ChuckHend:
Also a bug in pgmq-rs, should return False when queue already exists.

ChuckHend:
> Also a bug in pgmq-rs, should return False when queue already exists.

decided to ignore this and return unit to be consistent with all create functions

PR #579, opened 2026-08-23
feat!: Add `Queue` trait methods to create unlogged & partitioned queues

spencewenski:
This PR adds `create_unlogged`, `create_partitioned`, and `convert_archive_partitioned` methods to the `Queue` trait and adds implementations for all the sql drivers we support (sqlx, diesel, rust-postgres).

This PR also
- Adds the `partition_interval` and `retention_interval` parameters to the `PGMQueueExt#create_partitioned` method (breaking change).
- Adds `list_queues` and `queue_metadata` methods to `Queue` because these are useful for testing the `create_*` methods added in this PR, and they're helpful to check if a queue already exists. For example, `queue_metadata` is now used in `PGMQueueExt` to check if a queue exists instead of the previous `EXISTS` SQL query. This is a breaking change because it changes the constraints of the executor (`E`) type parameter of these methods.
- Changes the logic in `PGMQueueExt#create_partitioned` to check if a queue exists already: uses the `queue_metadata` function instead of inspecting a table owned by `pg_partman`
- Removes the `bool` return value of the `create*` methods as we don't believe this information is very relevant/valuable, and we don't need to check if the queue exists already for the `create` and `create_unlogged` methods.
- Updates the relevant methods in `PGMQueueExt` to use the sqlx implementation of `Queue`.


pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -277,45 +257,40 @@ impl PGMQueueExt {
...
+            retention_interval,
+            &self.connection,
+        )
+        .await
     }
 
     pub async fn convert_archive_partitioned_with_cxn<
         'c,
-        E: sqlx::Executor<'c, Database = Postgres> + std::marker::Copy,
+        E: sqlx::Executor<'c, Database = Postgres>,
```
spencewenski, line 311: I'm open to changing this method to use the implementation from the `crate::queue::sqlx` module like I've done for many of the other methods, but in that case I would change the `partition_interval` and `retention_interval` parameters to be non-optional.

pgmq-rs/src/pg_ext/mod.rs
```diff
@@ -366,6 +341,38 @@ impl PGMQueueExt {
         .await
     }
 
+    /// Idempotently create a queue using the provided `create_queue` function. Starts a
+    /// transaction, acquires a lock on the provided `queue_name`, checks whether the queue exists
+    /// already, and only runs `create_queue` if the queue does not already exist. This is useful
+    /// to avoid conflicts due to race conditions if, e.g., multiple processes try to create
+    /// the same queue in parallel.
+    async fn create_queue_idempotent<'c, E>(
```
ChuckHend: Can you confirm that the race condition still exists? We implemented [locking ](https://github.com/pgmq/pgmq/blob/c41b93adc4a93914339bde0ec3792311191f9e73/pgmq-extension/sql/pgmq.sql#L1186) within the server side create_queue() functionality.

spencewenski: It appears you're correct that the `pgmq.create*` SQL functions do not have this race condition anymore thanks to acquiring the lock. However, the `PGMQueueExt#create*` Rust functions do have a race condition related to their boolean return value -- if they don't acquire the queue lock, two invocations may return `true` even though only one actually created the queue.

Would we want to remove the boolean return value? I noticed there's an old issue related to this: https://github.com/pgmq/pgmq/issues/127

ChuckHend (approved):

pgmq-rs/src/pg_ext/mod.rs
ChuckHend: Yeah, I think we could remove it. Like @v0idpwn mentioned in #127, it is not that relevant whether the queue is new or not.

spencewenski: Actually, I discovered `pgmq.create_partitioned` can not be called twice for the same queue, so we need to check if the queue exists before calling `create_partitioned`. That must be what this comment is referring to: https://github.com/pgmq/pgmq/blob/c41b93adc4a93914339bde0ec3792311191f9e73/pgmq-rs/src/pg_ext/mod.rs#L290-L291

This isn't just in the case of a race condition -- even sequential calls to `create_partitioned` raise an error:

```
called `Result::unwrap()` on an `Err` value: TokioPostgresError(Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(EP0001), message: "duplicate key value violates unique constraint \"part_config_parent_table_pkey\"\nCONTEXT: SQL statement \"INSERT INTO public.part_config (\n        parent_table\n        , partition_type\n        , partition_interval\n        , control\n       , premake\n        , constraint_cols\n        , automatic_maintenance\n        , jobmon\n        , template_table\n        , inherit_privileges\n        , default_table\n        , date_trunc_interval)\nVALUES (\n        p_parent_table\n        , p_type\n        , v_id_interval\n        , p_control\n        , p_premake\n        , p_constraint_cols\n        , p_automatic_maintenance\n        , p_jobmon\n        ,v_template_schema||'.'||v_template_tablename\n        , v_inherit_privileges\n        , p_default_table\n        , p_date_trunc_interval)\"\nPL/pgSQL function create_parent(text,text,text,text,text,integer,text,boolean,text,text[],text,boolean,text) line 499 at SQL statement\nSQL statement \"\n    SELECT public.create_parent(\n      p_parent_table := 'pgmq.q_queue',\n      p_control := 'msg_id',\n      p_interval := '1000',\n      p_type := case\n        when pgmq._get_pg_partman_major_version() = 5 then 'range'\n        else 'native'\n      end\n    )\n    \"\nPL/pgSQL function pgmq.create_partitioned(text,text,text) line 33 at EXECUTE\nDETAIL: Key (parent_table)=(pgmq.q_queue) already exists.\nHINT: ", detail: None, hint: None, position: None, where_: Some("PL/pgSQL function create_parent(text,text,text,text,text,integer,text,boolean,text,text[],text,boolean,text) line 647 at RAISE\nSQL statement \"\n    SELECT public.create_parent(\n      p_parent_table := 'pgmq.q_queue',\n      p_control := 'msg_id',\n      p_interval := '1000',\n      p_type := case\n        when pgmq._get_pg_partman_major_version() = 5 then 'range'\n        else 'native'\n      end\n    )\n    \"\nPL/pgSQL function pgmq.create_partitioned(text,text,text) line 33 at EXECUTE"), schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("pl_exec.c"), line: Some(3897), routine: Some("exec_stmt_raise") }) })
```

So, I will keep the logic from the "create_idempotent" function, but since it's only needed in `create_partitioned`, I'll move the logic to that function.

ChuckHend: Oh good call. I forgot about the pg_partman part having concurrency issues on create.

.github/workflows/crate_ci.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           profile: minimal
           toolchain: stable
           components: rustfmt, clippy
-      - uses: taiki-e/install-action@9adcff13829681f1e2f4a678b775043ab5c7fc2c
+      - uses: taiki-e/install-action@37f7c5781271959fb65b6b35224e28652ff2b63d
```
spencewenski, line 31: Updating `taiki-e/install-action` was required to resolved the following error:

```
cargo semver-checks
    Building pgmq v0.34.0-alpha.6 (current)
       Built [  56.527s] (current)
     Parsing pgmq v0.34.0-alpha.6 (current)
error: unsupported rustdoc format v60 for file: /home/runner/work/pgmq/pgmq/pgmq-rs/target/semver-checks/local-pgmq-0_34_0_alpha_6-default-b790a1eae5f59d14/target/doc/pgmq.json
(supported formats are v56, v57)
make: *** [Makefile:53: check.semver] Error 1
```

https://github.com/pgmq/pgmq/actions/runs/33137733436/job/98744876303

pgmq-rs/src/pg_ext/mod.rs
ChuckHend: I think having consistent across the project would be better. I'm fine either way though.

spencewenski: Yeah, I agree. Will update.

PR #585, opened 2026-09-05
fix: partitioned queues silently overrun premake and cannot recover

### LINKED ISSUE #583: A burst past the premake runway silently fills the default partition and permanently breaks run_maintenance for that queue
### What happens

A partitioned queue that receives more messages than the premake runway between two maintenance runs puts the overflow in the default partition. Sends keep succeeding, reads keep working, nothing is logged — and from that moment `run_maintenance` for that queue **fails permanently**, so no new partitions are created ahead and retention stops dropping anything.

Reproduced on `ghcr.io/pgmq/pg17-pgmq:latest`, pg_partman 5.x, using only documented API:

```sql
CREATE EXTENSION pgmq CASCADE;
CREATE EXTENSION pg_partman;

-- small numbers so the runway is easy to exhaust; the shape is the same at any scale
SELECT pgmq.create_partitioned('burst', '10', '100');
-- partitions created: q_burst_p0 … q_burst_p40, plus q_burst_default → runway is msg_id < 50

SELECT pgmq.send('burst', '{"n":1}') FROM generate_series(1, 100);
-- all 100 succeed

SELECT count(*), min(msg_id), max(msg_id) FROM pgmq.q_burst_default;
--  51 | 50 | 100
```

Then maintenance, which is what the BGW runs on a schedule:

```sql
SELECT run_maintenance('pgmq.q_burst');
-- ERROR: updated partition constraint for default partition "q_burst_default"
--        would be violated by some row
```

It cannot recover on its own: the rows in the default partition are the reason the partitions covering them cannot be attached, and nothing moves them out. Every later message lands in the default too, because the runway never grows again.

### Why the runway is what it is

`pgmq.create_partitioned` calls `create_parent(...)` without `p_premake` and then updates `part_config` for `retention` only, so premake stays at pg_partman's default of 4 — about five intervals of runway — and pgmq exposes no way to change it.

With the defaults the docs describe (`partition_interval := '10000'`, `pg_partman_bgw.interval = 60`), that runway is ~50,000 msg_ids per 60 seconds. Sustained throughput above roughly 800 messages/second, or any single burst larger than the runway, crosses it. That is inside the range pgmq is built for, which is what makes this worth reporting rather than a curiosity.

The same applies on the time-based path when a queue's `partition_interval` is short relative to how long the BGW is delayed or paused.

### pg_partman's own recovery does not work here

`partition_data_proc` is the documented way to move rows out of a default partition. On a pgmq queue it fails:

```sql
CALL partition_data_proc('pgmq.q_burst', p_interval := '10');
-- ERROR: cannot insert a non-DEFAULT value into column "msg_id"
```

`msg_id` is `GENERATED ALWAYS AS IDENTITY`, and the move inserts it explicitly without `OVERRIDING SYSTEM VALUE`. So the standard remedy is unavailable precisely on the tables that need it.

### A recovery that does work

Tested on the reproduction above — 100 messages in, 100 messages out, `run_maintenance` healthy afterwards:

```sql
BEGIN;
ALTER TABLE pgmq.q_burst DETACH PARTITION pgmq.q_burst_default;
SELECT create_partition_id('pgmq.q_burst',
       ARRAY(SELECT DISTINCT (msg_id / 10) * 10 FROM pgmq.q_burst_default));   -- 10 = partition_interval
INSERT INTO pgmq.q_burst OVERRIDING SYSTEM VALUE SELECT * FROM pgmq.q_burst_default;
TRUNCATE pgmq.q_burst_default;
ALTER TABLE pgmq.q_burst ATTACH PARTITION pgmq.q_burst_default DEFAULT;
COMMIT;
```

A warning worth having in writing, because it is the obvious thing to try first and it destroys messages: detaching the default and then calling `run_maintenance` does **not** create the covering partitions (maintenance works from the parent, which no longer sees those rows), so the `INSERT` fails with `no partition of relation "q_burst" found for row` — and if the statements are not in one transaction, a following `DELETE`/`TRUNCATE` of the default partition throws the messages away. I did exactly that on the first attempt and lost 51 of 100. The partitions have to be created explicitly, by id, before the rows move.

### What might help

Your call which of these is worth doing, if any:

- **Expose `p_premake`** on `pgmq.create_partitioned`, so a high-throughput queue can be given a runway that matches it. Today it cannot be set through pgmq at all, and `part_config` is already being updated in that function, so there is a natural place for it.
- **Say in `docs/partitioned-queues.md`** how `partition_interval`, premake and `pg_partman_bgw.interval` combine into a throughput ceiling, and what happens when it is crossed. The page documents each setting well; it is the interaction that bites.
- **Point operators at `check_default()`** — pg_partman ships it for exactly this, and a non-empty default partition on a pgmq queue means maintenance is already broken, not merely that some rows are misfiled.
- Possibly **fail loudly** instead: pgmq could omit the default partition, in which case an overflowing send errors immediately rather than silently starting an unrecoverable state. That is a product decision about which failure you prefer, and I have no view on it from outside.

---

Disclosure: I maintain a PostgreSQL partition-lifecycle library, which is why I was reading partitioning code across projects. Nothing here involves it — pgmq's queue operations are server-side SQL and should stay that way. This is a report about the pg_partman wiring and the docs.

AlexeyShalaev:
Closes #583. This is the list from your comment, in one PR.

**`pgmq.create_partitioned()` takes `premake`** (default 4, which is pg_partman's own, so nothing changes for existing calls) and passes it to both `create_parent` calls, queue table and archive table. Values below 1 are rejected with a plain message. The three-argument function is dropped in the migration rather than kept alongside, so `create_partitioned('q', '10', '100')` still resolves to exactly one function.

**`msg_id` on partitioned queues is `GENERATED BY DEFAULT AS IDENTITY`.** With that, `partition_data_proc` works as the recovery path, which is a much better story than the hand-written transaction I posted on the issue. The migration flips existing partitioned queues in place, following the loop from the 1.9.0 to 1.10.0 migration.

**`pgmq.metrics()` reports `default_partition_length`**, appended last: the row count of the queue's default partition, `null` for queues that aren't partitioned. Zero is the only healthy value. `metrics_all()` picks it up through `metrics()`.

**On calling `partition_data_proc` from the migration**, which you were unsure about: it can't be done. The procedure commits as it goes, and inside a transaction block it fails with `invalid transaction termination`, which is what `ALTER EXTENSION UPDATE` is. The function form `partition_data_id()` does run inside a transaction, but it would move an unbounded amount of data in a single transaction in the middle of an extension upgrade, which I don't think anyone wants to discover at upgrade time. So the migration changes the identity and nothing else, and recovery is two documented statements:

```sql
CALL partman.partition_data_proc('pgmq.q_my_queue');
SELECT partman.run_maintenance('pgmq.q_my_queue');
```

Docs: `partitioned-queues.md` gets the `premake` parameter and a "Runway and overrun" section covering the throughput ceiling, why crossing it is silent, monitoring via the new metric, recovery, and the `BY DEFAULT` trade-off (an explicit `msg_id` from something bypassing the API can collide with a later `send()`; pgmq itself never supplies one). `functions.md` and `types.md` updated to match.

Tested on `ghcr.io/pgmq/pg17-pgmq` with pg_partman 5.1.0:

- `pg_regress base` on a fresh 1.12.1 install passes with the regenerated expected output, which changes in exactly three places: two `CONTEXT` lines that now show the four-argument signature, and the new test section. There is one pre-existing diff in `base` unrelated to this change, a `msg_id 5/6` ordering flip in a query without `ORDER BY` around line 632; it reproduces identically on unmodified 1.12.0, so I left the expected file's ordering alone.
- Upgrade path: install 1.12.0, create a partitioned queue, overrun it (51 rows in the default partition), `ALTER EXTENSION pgmq UPDATE TO '1.12.1'`. Afterwards `identity_generation` is `BY DEFAULT`, `metrics()` reports 51, the old three-argument call and the Rust client's named-argument call both work, `premake => 8` lands in `part_config`, `premake => 0` is rejected, `partition_data_proc` moves the 51 rows and `run_maintenance` is healthy again.

Left alone on purpose: `convert_archive_partitioned()` calls `create_parent` too and could take `premake` the same way. I kept this PR to what the issue covers; happy to do that as a follow-up if you want it. The Rust client needs no change: it calls `create_partitioned` with named arguments and selects metrics columns by name.


docs/partitioned-queues.md
```diff
@@ -27,11 +29,34 @@ in `postgresql.conf`. Below are the default configuration values set in pgmq doc
...
+### Recovery
+
+Move the messages out of the default partition into the partitions they belong to, then let maintenance catch up. `pg_partman` provides the procedure for this; run it outside a transaction block, because it commits as it goes:
+
+```sql
+CALL partman.partition_data_proc('pgmq.q_my_queue');
+SELECT partman.run_maintenance('pgmq.q_my_queue');
+```
+
+`msg_id` on partitioned queues is declared `GENERATED BY DEFAULT AS IDENTITY` (rather than `ALWAYS`) precisely so that this move can reinsert messages with their existing ids. The trade-off is that PostgreSQL no longer rejects an explicit `msg_id` in an `INSERT`. `pgmq` never supplies one, but if something writes to a queue table directly and provides an id ahead of the sequence, a later `pgmq.send()` will collide with it. Go through the `pgmq` API.
```
ChuckHend, line 65: 👍

pgmq-extension/sql/pgmq--1.12.0--1.12.1.sql
```diff
@@ -0,0 +1,272 @@
...
+-- partition by inserting them with their existing msg_id, which
+-- GENERATED ALWAYS refuses. pgmq's own writers never supply msg_id, so this
+-- changes nothing for them.
+DO $$
+DECLARE
+    queue_record RECORD;
+    qtable TEXT;
+BEGIN
+    FOR queue_record IN SELECT queue_name FROM pgmq.meta WHERE is_partitioned LOOP
+        qtable := 'q_' || queue_record.queue_name;
```
ChuckHend: Should help if there are any issues with formatting.

```suggestion
        qtable := pgmq.format_table_name(queue_record.queue_name, 'q');
```

pgmq-extension/sql/pgmq.sql
```diff
@@ -873,7 +874,16 @@ DECLARE
...
     qtable TEXT := pgmq.format_table_name(queue_name, 'q');
+    default_partition TEXT := qtable || '_default';
+    default_partition_length BIGINT;
 BEGIN
+    -- Only partitioned queues have a default partition. Messages in it have no
+    -- partition of their own, which means pg_partman maintenance is failing for
+    -- this queue; a non-zero value here is the signal to act on.
+    IF to_regclass(FORMAT('pgmq.%I', default_partition)) IS NOT NULL THEN
+        EXECUTE FORMAT('SELECT count(*) FROM pgmq.%I', default_partition) INTO default_partition_length;
+    END IF;
```
ChuckHend: I think we need to use an estimate here. If there's a lot of records spilled over into the default partition it could get quite slow. This is also similar to how pg_partman's check_default() is set up.

```suggestion
    IF to_regclass(FORMAT('pgmq.%I', q_default_partition)) IS NOT NULL THEN
        SELECT COALESCE(SUM(GREATEST(c.reltuples, 0))::bigint, 0)
        INTO default_partition_length
        FROM pg_class c
        WHERE c.oid IN (
            to_regclass(FORMAT('pgmq.%I', q_default_partition)),
            to_regclass(FORMAT('pgmq.%I', a_default_partition))
        );
    END IF;
```

pgmq-extension/sql/pgmq--1.12.0--1.12.1.sql
```diff
@@ -0,0 +1,272 @@
...
+CREATE TYPE pgmq.metrics_result AS (
+    queue_name text,
+    queue_length bigint,
+    newest_msg_age_sec int,
+    oldest_msg_age_sec int,
+    total_messages bigint,
+    scrape_time timestamp with time zone,
+    queue_visible_length bigint,
+    default_partition_length bigint
+);
```
ChuckHend: We should be able to just modify the type in place.

```suggestion
ALTER TYPE pgmq.metrics_result ADD ATTRIBUTE default_partition_length bigint;
```

pgmq-extension/sql/pgmq--1.12.0--1.12.1.sql
```diff
@@ -0,0 +1,272 @@
...
+DECLARE
+    row_name RECORD;
+    result_row pgmq.metrics_result;
+BEGIN
+    FOR row_name IN SELECT queue_name FROM pgmq.meta LOOP
+        result_row := pgmq.metrics(row_name.queue_name);
+        RETURN NEXT result_row;
+    END LOOP;
+END;
+$$ LANGUAGE plpgsql;
```
ChuckHend: If we aren't dropping the `pgmq.metrics_result TYPE` then I don't think we'll have to drop and replace pgmq."metrics_all"() either.
```suggestion
```

ChuckHend (commented):
This looks great! Thanks for producing this so quickly. Docs and tests are good to go. I left a few inline comments related to the migration and monitoring more closely to how pg_partman [handles it internally](https://github.com/pgpartman/pg_partman/blob/c737cc1d32a46a37120aa691f7dd0bce6ad07660/sql/functions/check_default.sql).

pgmq-extension/sql/pgmq.sql
AlexeyShalaev: Took this as is, with the archive's default partition included, and the docs now call it an estimate. One property of `reltuples` worth knowing for this particular number: it only moves when autovacuum or `ANALYZE` visits the default partition. On the way in that's fine, a spilled partition keeps growing and autoanalyze picks it up within a naptime. On the way out it lags: after `partition_data_proc` the partition is empty but `reltuples` still reports the old count (51 in my run), and for a small spill, under the autovacuum thresholds, autovacuum never gets around to it. So the recovery recipe in the docs now ends with `ANALYZE` on the default partition, and the tests do the same before reading the value.

If you'd rather have something that never lags: the non-exact branch of `check_default()` is actually a `LIMIT 1` probe rather than `reltuples`, instant, but it only says that something is there, not how much you're about to move. The estimate seemed the more useful number for an operator, so I left it as you wrote it.

pgmq-extension/sql/pgmq--1.12.0--1.12.1.sql
AlexeyShalaev: Done. That is also what create_partitioned() builds the table from, so the flip now derives the name the same way the table got it.

pgmq-extension/sql/pgmq--1.12.0--1.12.1.sql
AlexeyShalaev: Done. I should have looked at what 1.4.5 → 1.5.0 did for queue_visible_length, it is the same change.

pgmq-extension/sql/pgmq--1.12.0--1.12.1.sql
AlexeyShalaev: Removed. With the type altered in place, metrics() is a CREATE OR REPLACE and metrics_all() needs no touching.

docs/partitioned-queues.md
AlexeyShalaev: This paragraph moved a little since your 👍: it now says the value is an estimate, covers the archive's default partition too, and the recovery recipe ends with an ANALYZE so the metric drops to zero right away.

ChuckHend (approved):
Awesome, thanks again for raising the issue and contributing!

PR #593, opened 2026-09-17
fix: insert notifications never fire for a queue name that is not all lowercase

Dev-next-gen:
I was looking at `pgmq.notify_queue_listeners()` and noticed it rebuilds the queue name from the table the trigger sits on:

```sql
queue_name_extracted := substring(TG_TABLE_NAME from 3);
```

That reconstruction cannot work, because `format_table_name()` lowercases. A queue created as `MyQueue` lives in `pgmq.q_myqueue`, so stripping the prefix yields `myqueue`, while `pgmq.notify_insert_throttle.queue_name` holds `MyQueue` (it is a foreign key to `pgmq.meta`, which stores the name as given). The throttle `UPDATE` matches no row, `updated_count` stays 0, and the `PG_NOTIFY` it guards is never reached. Nothing is logged and no error is raised: `enable_notify_insert()` succeeds, `list_notify_insert_throttles()` shows the queue as enabled, and the consumer simply waits forever. No listener can work around it, since no notification is sent at all.

On `ghcr.io/pgmq/pg17-pgmq` with the released 1.13.0, listening on both spellings of the channel:

```
SELECT pgmq.create('MyQueue');
SELECT pgmq.enable_notify_insert('MyQueue', 0);
LISTEN "pgmq.q_MyQueue.INSERT";
LISTEN "pgmq.q_myqueue.INSERT";
SELECT pgmq.send('MyQueue', '{"a":1}'::jsonb);

 queue_name | notified
------------+----------
 MyQueue    | f          -- no notification on either channel
 lowqueue   | t          -- same steps, all-lowercase name
```

`enable_notify_insert()` knows the real queue name, so I pass it to the trigger function as an argument instead of having the function guess it back from the table name. The throttle row is then found, and the channel carries the queue name as given — which is what `pgmq.q_<queue_name>.INSERT` already meant for every name that worked, since those were all lowercase. For an all-lowercase queue the channel string is byte-identical to today's, so nothing that works now changes; `pgmq-extension/client-tests/basic_notify_test.py` and the Rust client's `queue_name_to_insert_notification_channel_name` both already build the channel from the raw queue name, and they now hold for mixed-case names too. I updated the note in `functions.md` that described the channel as the internal table name, which was the ambiguous part.

Triggers created by the previous `enable_notify_insert` carry no argument, so `TG_ARGV[0]` is null and the function falls back to the table name — same behaviour and same channel as before for them. Rather than leave existing installs waiting for someone to call `enable_notify_insert` again, the migration re-creates those triggers with the queue name attached, following the loop from the 1.9.0 to 1.10.0 migration; throttle intervals are read from `notify_insert_throttle`, not reset. This is a fix with no API change, so I targeted a patch release, 1.13.1.

Verified on `ghcr.io/pgmq/pg17-pgmq` with pg_partman 5.1.0:

- New regress case in `base`: `notified` is `f` on unmodified 1.13.0 and `t` after the change. I assert that `last_notified_at` moved off the epoch rather than catching the notification itself, because the psql async-notification line carries a backend PID and would not be reproducible.
- `pg_regress base fifo_tests fifo_rr_tests fifo_per_group_tests topic_tests`: all pass with the regenerated expected output. Running the unmodified test files against the patched extension leaves exactly one diff, the `msg_id 5/6` ordering flip around line 632 of `base.out` in a query without `ORDER BY`; it reproduces identically on unmodified 1.13.0 in the same container, so it is pre-existing and I left the expected file's ordering alone.
- Upgrade in place: install 1.13.0, create `MyQueue`, enable notifications, send (nothing arrives), `ALTER EXTENSION pgmq UPDATE TO '1.13.1'`, send again — the notification arrives on `pgmq.q_MyQueue.INSERT` and the throttle interval is unchanged.
- `pg_get_functiondef` for every `pgmq` function is identical between a fresh 1.13.1 install and a 1.13.0 install upgraded to 1.13.1, so the migration does not diverge from the full script.
- Fallback path: re-created an argument-less trigger by hand on a lowercase queue against the patched function; it still notifies on the same channel.

One thing you may want to decide differently: I made the channel follow the queue name, which is the shape both client test suites assume. The other option is to keep the channel on the lowercased table name and fix only the throttle lookup, which would then require the Rust and Python helpers to lowercase. I went with the client-facing convention because it leaves every currently working channel name untouched.

While reading around this I also found that two queues whose names differ only in case share a single physical table — `create('CaseTest')` then `create('casetest')` gives two `pgmq.meta` rows for one `pgmq.q_casetest`, and messages sent to one are read from the other. That looked like a separate question about whether such names should be rejected, so I left it out of this PR; happy to open an issue for it if useful.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


ChuckHend:
Hey @Dev-next-gen, I was able to reproduce. Thanks for the PR, this is a good fix.

ChuckHend (approved):

